PR #29, opened 2021-11-15
Add [CallerArgumentExpression] to Guard APIs

### LINKED ISSUE #3: Add [CallerArgumentExpression] support to Guard APIs
## Describe the problem this feature would solve

C# 10 introduces support for the `[CallerArgumentExpression]` aqttribute, which lets the compiler access the argument of a given expression as input, without the user needing to type it out manually. This can be particularly beneficial in the `Guard` APIs, as it'd allow callers to avoid to specify the name of the parameter being checked in most cases. For instance:

```csharp
// Before
Guard.IsNotNull(someVariable, nameof(someVariable));

// After
Guard.IsNotNull(someVariable);
```

And this would result in the same generated code. It would also still be possible to manually specify the inputs where needed.

## Describe the solution

The complete diff is too big, but really it's just a matter of doing this same change for all APIs:

```diff
 namespace CommunityToolkit.Diagnostics
 {
     public static class Guard
     {
-        public static void IsNotNull<T>(T? value, string name);
+        public static void IsNotNull<T>(T? value, [CallerArgumentExpression("value")] string name = "");
     }
}
```

Sergio0694:
**Closes #3** 

This PR adds support for `[CallerArgumentExpression]` to all `Guard` APIs.

### DIFF
--- a/CommunityToolkit.Diagnostics/Attributes/CallerArgumentExpressionAttribute.cs
+++ b/CommunityToolkit.Diagnostics/Attributes/CallerArgumentExpressionAttribute.cs
@@ -1,0 +1,31 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if !NET5_0_OR_GREATER
+
+namespace System.Runtime.CompilerServices;
+
+/// <summary>
+/// An attribute that allows parameters to receive the expression of other parameters.
+/// </summary>
+/// <remarks>Internal copy from the BCL attribute.</remarks>
+[AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
+internal sealed class CallerArgumentExpressionAttribute : Attribute
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CallerArgumentExpressionAttribute"/> class.
+    /// </summary>
+    /// <param name="parameterName">The condition parameter value.</param>
+    public CallerArgumentExpressionAttribute(string parameterName)
+    {
+        ParameterName = parameterName;
+    }
+
+    /// <summary>
+    /// Gets the parameter name the expression is retrieved from.
+    /// </summary>
+    public string ParameterName { get; }
+}
+
+#endif
--- a/CommunityToolkit.Diagnostics/Generated/Guard.Collection.g.cs
+++ b/CommunityToolkit.Diagnostics/Generated/Guard.Collection.g.cs
@@ -24,7 +24,7 @@ namespace CommunityToolkit.Diagnostics
         /// <param name="name">The name of the input parameter being tested.</param>
         /// <exception cref="ArgumentException">Thrown if the size of <paramref name="span"/> is != 0.</exception>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static void IsEmpty<T>(Span<T> span, string name)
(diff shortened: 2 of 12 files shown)


CommunityToolkit.Diagnostics/Attributes/CallerArgumentExpressionAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+// See the LICENSE file in the project root for more information.
+
+#if !NET5_0_OR_GREATER
+
+namespace System.Runtime.CompilerServices;
+
+/// <summary>
+/// An attribute that allows parameters to receive the expression of other parameters.
+/// </summary>
+/// <remarks>Internal copy from the BCL attribute.</remarks>
```
michael-hawker, line 12: This basically just makes the attribute a no-op to compile for .NET Standard 2.0, eh?

Do we need to worry about if folks drop the argument on non .NET 5+ platforms? Is that something we can analyze or provide info on somewhere?

michael-hawker (approved):

CommunityToolkit.Diagnostics/Attributes/CallerArgumentExpressionAttribute.cs
Sergio0694: We don't need to worry, this feature will work regardless of target runtime as long as you're using VS2022+, since it's purely a Roslyn feature, it's just syntax sugar really. This means it will also work on eg. .NET Standard 2.0, UWP, etc. just fine. This attribute is just needed because Roslyn looks it up and matches it by name when compiling code, but it doesn't actually need to be part of the runtime itself (as in, it's like the nullability attributes that can be polyfilled in the same way, not like eg. `[StackTraceHidden]` which needs explicit runtime support and is matched by identity and not name).

XAML-Knight (approved):

CommunityToolkit.Diagnostics/Attributes/CallerArgumentExpressionAttribute.cs
```diff
@@ -0,0 +1,31 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if !NET5_0_OR_GREATER
```
Youssef1313, line 5: Per the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callerargumentexpressionattribute?view=net-6.0), the attribute existed in .NET Core 3.0, but had no effect. I think it shouldn't be redefined for .NET Core 3.0, .NET Core 3.1

michael-hawker: We're referencing .NET Standard 2.0 which does not include this attribute, so we still need it? Or are you saying we should make a call out specific for .NET Core 3+ so we don't duplicate it on that runtime?

Sergio0694: Yeah we don't have a .NET Core 3.1 target, and I don't think it's worth it adding one. Note that our polyfill copies for this attribute are internal, so even if you were to use the MVVM Toolkit in a .NET Core 3.1 project, there would not be an issue: Roslyn would just use the copy defined here to match the attribute used on the `Guard` APIs, and you'd keep using the public one from the BCL anywhere else, there shouldn't be any conflicts. It's actually not uncommon to have duplicate internal copies of attributes, as many libraries targeting older frameworks do that (we do the same as well), and due to how NuGet package resolution works it's very possible you'll end up with one of those assemblies when using a newer target anyway 👍

Issue #46, opened 2021-11-30
Investigate AppContext switches for opt-in/opt-out features


Sergio0694:
This is a tracking issue about experimenting with `AppContext` switches to customize the MVVM Toolkit.

## Rationale

There are some proposed features (eg. #28) that cannot really be implemented in a transparent and pay-for-play with the current setup. They'd end up either requiring a separate type, or adding overhead for everyone. Similarly, some existing features, such as the `INotifyPropertyChanging` support in `ObservableObject` are not really needed by everyone, and it would be nice to have a way of disabling them. Obviously the interface cannot be removed at runtime, but it could just be not functional if desired.

## The idea

We could experiment with adding a bunch of `AppContext` switches (see [`AppContext.SetSwitch`](https://docs.microsoft.com/en-us/dotnet/api/system.appcontext.setswitch) and [`AppContext.TryGetSwitch`](https://docs.microsoft.com/en-us/dotnet/api/system.appcontext.trygetswitch)) with names that we would then document, that would enable users to manually opt-in and out of specific features. The default values would remain consistent with the behavior today, but this would enable more experienced users to further customize the behavior of the MVVM Toolkit to suit their needs. Those switches would be cached into `static readonly bool` fields, so that checking them would be virtually free on modern runtimes (eg. when using tiered JIT, which is the default on CoreCLR, such fields would just become constants and be inlined when the method is recompiled at runtime, so the conditional branch would just vanish completely).

## Feature items

- [ ] Dispatcher-away invoke for property changed/changing events
- [ ] Opt-out switch for `INotifyPropertyChanging`
- [ ] ..?

michael-hawker:
For `INotifyPropertyChanging` - Is there a big advantage to checking the bool and not firing the event if there's no handlers registered?

Could this be enabled better with the Source Generator support by not inheriting from `ObservableObject` and just using the `INotifyPropertyChanged` attribute? Could the `ObservableObject` generator attribute have a property for this to change the generated version as well?

Sergio0694:
> "Is there a big advantage to checking the bool and not firing the event if there's no handlers registered?"

Yes, calling `SetProperty` becomes about twice as fast and allocates half (as it's no longer also creating the `PropertyChangingEventArgs` object, which is spawned regardless of whether there are any handlers for the event).

Here's some numbers (benchmark is just doing `++` on an observable property that uses `SetProperty`):

[...]

PR #48, opened 2021-11-30
Enable INotifyPropertyChanging AppContext switch

Sergio0694:
**Contributes to #46**

This PR adds a new "MVVMTOOLKIT_DISABLE_INOTIFYPROPERTYCHANGING" switch that disables `INotifyPropertyChanging` in `ObservableObject` and inherited classes. This needs to be set through `AppContext.SetSwitch` right at startup.

**NOTE:** this is a draft as this branch is based on top of #47. Rebasing and switch to non-draft once that's merged.

### DIFF
--- a/tests/CommunityToolkit.Mvvm.DisableINotifyPropertyChanging.UnitTests/Test_DisableINotifyPropertyChanging.cs
+++ b/tests/CommunityToolkit.Mvvm.DisableINotifyPropertyChanging.UnitTests/Test_DisableINotifyPropertyChanging.cs
@@ -1,0 +1,60 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.ComponentModel;
+using CommunityToolkit.Mvvm.ComponentModel;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace CommunityToolkit.Mvvm.UnitTests.Configuration;
+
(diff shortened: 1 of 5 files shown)


XAML-Knight (approved):

tests/CommunityToolkit.Mvvm.DisableINotifyPropertyChanging.UnitTests/Test_DisableINotifyPropertyChanging.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.ComponentModel;
+using CommunityToolkit.Mvvm.ComponentModel;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace CommunityToolkit.Mvvm.UnitTests.Configuration;
+
+[TestClass]
+public class Test_DisableINotifyPropertyChanging
+{
+    static Test_DisableINotifyPropertyChanging()
```
Arlodotexe, line 15: Shouldn't this use `[ClassInitialize]`, and reset the switch on `[ClassCleanup]` so it can't affect other tests?

Sergio0694: The switch cannot be reset, as it's used to set a static read-only field 😉
This test file is in a separate, dedicated unit test project specifically to test this. Look at the file oath, you'll see it's not in the same unit test project as the others 🙂

Arlodotexe (approved):

PR #59, opened 2021-12-10
Optimize StrongReferenceMessenger

Sergio0694:
This PR brings some of the optimizations and improvements from #44 to `StrongReferenceMessenger` too.
Specifically, this includes:

- Using double-dispatch through `MessageHandlerDispatcher` to avoid aliasing delegate types during broadcast.
- Type specialization for `Unit` (used when sending messages through the default channel)
- Specialization for `IRecipient<TMessage>` to avoid the double delegate dispatch
- Other general tweaks and code refactoring

## Benchmarks

| Method |     Mean |     Error |    StdDev | Ratio | Allocated |
|------- |---------:|----------:|----------:|------:|----------:|
|   main | 5.586 ms | 0.0317 ms | 0.0264 ms |  1.00 |         - |
|     **PR** | **4.013 ms** | 0.0074 ms | 0.0058 ms |  **0.71** |         - |

Overall this results in a ~30% speedup 🚀

### DIFF
--- a/CommunityToolkit.Mvvm/Messaging/StrongReferenceMessenger.cs
+++ b/CommunityToolkit.Mvvm/Messaging/StrongReferenceMessenger.cs
@@ -122,7 +122,9 @@ public sealed class StrongReferenceMessenger : IMessenger
 
                 Recipient key = new(recipient);
 
-                return mapping.ContainsKey(key);
+                return
+                    mapping.TryGetValue(key, out Dictionary2<TToken, object?>? handlers) &&
+                    handlers.ContainsKey(token);
             }
         }
     }
--- a/CommunityToolkit.Mvvm/Messaging/StrongReferenceMessenger.cs
+++ b/CommunityToolkit.Mvvm/Messaging/StrongReferenceMessenger.cs
@@ -46,23 +46,28 @@ public sealed class StrongReferenceMessenger : IMessenger
     // which holds the references from registered recipients to handlers. Mapping is used when the default channel is being
     // requested, as in that case there will only ever be up to a handler per recipient, per message type. In that case,
     // each recipient will only track the message dispatcher (stored as an object?, see notes below), instead of a dictionary
-    // mapping each TToken value to the corresponding dispatcher for that recipient. When a custom channel is used, the dispatchers
-    // are stored in a <TToken, object?> dictionary, so that each recipient can have up to one registered handler for a given token,
-    // for each message type. Note that the registered dispatchers are only stored as object references, as they can either be null
-    // or a MessageHandlerDispatcher.For<TRecipient, TMessage> instance. The first case happens if the handler was registered through
-    // an IRecipient<TMessage> instance, while the second one is used to wrap input MessageHandler<TRecipient, TMessage> instances.
-    // The MessageHandlerDispatcher.For<TRecipient, TMessage> instances will just be cast to MessageHandlerDispatcher when invoking it.
-    // This allows users to retain type information on each registered recipient, instead of having to manually cast each recipient
-    // to the right type within the handler (additionally, using double dispatch here avoids the need to alias delegate types).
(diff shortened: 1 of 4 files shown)


CommunityToolkit.Mvvm/Messaging/StrongReferenceMessenger.cs
```diff
@@ -24,43 +24,50 @@ public sealed class StrongReferenceMessenger : IMessenger
...
     // Dictionary2<Type2, IMapping> typesMap;
     // --------------------------------------------------------------------------------------------------------
-    // Each combination of <TMessage, TToken> results in a concrete Mapping<TMessage, TToken> type, which holds the references
-    // from registered recipients to handlers. The handlers are stored in a <TToken, MessageHandler<object, TMessage>> dictionary,
-    // so that each recipient can have up to one registered handler for a given token, for each message type.
-    // Note that the registered handlers are only stored as object references, even if they were actually of type
-    // MessageHandler<TRecipient, TMessage>, to avoid unnecessary unsafe casts. Each handler is also generic with respect to the
-    // recipient type, in order to allow the messenger to track and invoke type-specific handlers without using reflection and
-    // without having to capture the input handler in a proxy delegate, causing one extra memory allocations and adding overhead.
+    // Each combination of <TMessage, TToken> results in a concrete Mapping type (if TToken is Unit) or Mapping<Token> type,
```
Arlodotexe, line 45: Maybe look at splitting this up into multiple paragraphs to improve readability

CommunityToolkit.Mvvm/Messaging/WeakReferenceMessenger.cs
```diff
@@ -218,6 +218,12 @@ public void UnregisterAll(object recipient)
     public void UnregisterAll<TToken>(object recipient, TToken token)
         where TToken : IEquatable<TToken>
     {
+        // This method is never called with the unit type
+        if (typeof(TToken) == typeof(Unit))
+        {
+            throw new NotImplementedException();
```
Arlodotexe, line 224: Since `Unit` is internal, what's the reason for throwing? Comment doesn't make it clear
Also Is `NotImplementedException` the best exception to use here?

CommunityToolkit.Mvvm/Messaging/StrongReferenceMessenger.cs
Sergio0694: Done in bf0fab6fa3bf1e8c2336ce352ed276cc739cea07 👍

CommunityToolkit.Mvvm/Messaging/WeakReferenceMessenger.cs
Sergio0694: The reason is just that the rest of the method is not valid if `Unit` was used, so we might as well just have an explicit check to ensure that it is indeed never called. Since `Unit` is a value type, this type check will be compeltely removed by the compiler anyway when not used, so it will be completely free. There is no check being done at runtime here 🙂

Arlodotexe: Ahh so it'll never actually throw NotImplementedException at runtime. I assumed we avoid throwing this in a production app -- maybe we should change it to InvalidOperationException or NotSupportedException to set a good example?

Sergio0694: I think `NotImplementedException` is fine here. The exact exception type doesn't really matter much anyway, this is not an actual exception we'll ever be throwing. If that had been the case, then yeah I'd have opted for `NotSupportedException`. But in this case this path is actually, by design, not implemented. I've added more documentation in 17b046513ca51c093ccbc0b58f078011a6072a5a though, which should hopefully make the setup clearer and easier to understand, let me know if that works! 🙂

Arlodotexe (approved):

PR #60, opened 2021-12-13
Add null checks for parameters of public MVVM Toolkit APIs

Sergio0694:
This PR adds `null` checks throwing `ArgumentNullException`-s to all nullable parameters in public APIs in the MVVM Toolkit. This helps developers narrow down issues faster, and it also avoids state corruption in case an invalid input is given to an API, and an exception is throwin in the middle of some other operation, leading to an undefined state. As a result, behavior should now be properly defined for all possible inputs.


Arlodotexe (approved):

CommunityToolkit.Mvvm/ComponentModel/ObservableValidator.cs
```diff
@@ -629,14 +683,8 @@ protected internal void ValidateProperty(object? value, [CallerMemberName] strin
...
-    private bool TryValidateProperty(object? value, string? propertyName, out IReadOnlyCollection<ValidationResult> errors)
+    private bool TryValidateProperty(object? value, string propertyName, out IReadOnlyCollection<ValidationResult> errors)
     {
-        if (propertyName is null)
-        {
-            ThrowArgumentNullExceptionForNullPropertyName();
-        }
-
         // Add the cached errors list for later use.
         if (!this.errors.TryGetValue(propertyName!, out List<ValidationResult>? propertyErrors))
```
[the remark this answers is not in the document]
Sergio0694: Good point, fixed in 68b715f688bac3dac694d05770a9cd2ba6121f70 👍

Arlodotexe (approved):

tests/CommunityToolkit.Mvvm.UnitTests/Test_ArgumentNullException.Input.cs
```diff
@@ -0,0 +1,63 @@
...
+using CommunityToolkit.Mvvm.DependencyInjection;
+using CommunityToolkit.Mvvm.Input;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace CommunityToolkit.Mvvm.UnitTests;
+
+public partial class Test_ArgumentNullException
+{
+    [TestMethod]
+    public void Test_RelayCommand()
```
michael-hawker, line 15: The test name is only thing that shows up in devops, do we want to add null to the method name somewhere?

image

Sergio0694: Yup, I forgot to add the name of the containing type to these new tests, good catch! 😄
Fixed in 7f0c5a1919dbf5dd29d749c840569e8f0628dc1d.

XAML-Knight (approved):

PR #64, opened 2021-12-17
Add NullableExtensions.DangerousGetValueOrNullReference

### LINKED ISSUE #61: NullableExtensions.DangerousGetValueOrNullReference()
The existing `DangerousGetValueOrDefaultReference()` is useful, but for a lot of my interop code I need `null` when `HasValue` is false

I'm actually using a more dangerous version in my code base that returns a pointer, but @Sergio0694 says return a null `ref` is a better idea, and I agree. Can always `Unsafe.AsPointer()` from that.

```
public static unsafe ref T DangerousGetValueOrNullReference<T>(ref this T? maybe)
    where T : unmanaged
{
    return ref (maybe.HasValue ? ref maybe.DangerousGetValueOrDefaultReference() : ref Unsafe.NullRef<T>());
}
```

Sergio0694:
**Closes #61**

This PR adds the `NullableExtensions.DangerousGetValueOrNullReference` method.
It also adjusts the namespace for `NullableExtensions`, which was incorrect.

cc. @rickbrew

### API breakdown:

```csharp
namespace CommunityToolkit.HighPerformance
{
    public static class NullableExtensions
    {
        public static unsafe ref T DangerousGetValueOrNullReference<T>(ref this T? value)
            where T : struct;
    }
}
```

### DIFF
--- a/CommunityToolkit.HighPerformance/Extensions/NullableExtensions.cs
+++ b/CommunityToolkit.HighPerformance/Extensions/NullableExtensions.cs
@@ -12,7 +12,7 @@
 using System;
 using System.Runtime.CompilerServices;
 
-namespace CommunityToolkit.HighPerformance.Extensions;
+namespace CommunityToolkit.HighPerformance;
 
 /// <summary>
 /// Helpers for working with the <see cref="Nullable{T}"/> type.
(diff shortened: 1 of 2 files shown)


rickbrew (approved):
lgtm 😺👍🏼

XAML-Knight (approved):

PR #65, opened 2021-12-17
Fix RuntimeHelpers.IsReferenceOrContainsReferences polyfill

### LINKED ISSUE #32: Fix portable IsReferenceOrContainsReference<T> for pointer types
## Overview

The `IsReferenceOrContainsReferences(Type)` method we use to polyfill on .NET Standard < 2.1 has a bug that was inherited from the portable implementation originally in the BCL, which causes pointer types or types containing pointer types to be reported as being reference types. This is caused by `Type.IsValueType` actually returning `false` in this case (for... Reasons? 🤷).

https://github.com/CommunityToolkit/dotnet/blob/1a94430894333ca559b88f32665e6b82f36ebc87/CommunityToolkit.HighPerformance/Helpers/Internals/RuntimeHelpers.cs#L184

There should be an additional check for pointer types, and also a workaround for `IsValueType`.

Sergio0694:
**Closes #32**

This PR fixes the internal `RuntimeHelpers.IsReferenceOrContainsReferences<T>` polyfill for pointer types.

### DIFF
--- a/tests/CommunityToolkit.HighPerformance.UnitTests/Helpers/Internals/Test_RuntimeHelpers.cs
+++ b/tests/CommunityToolkit.HighPerformance.UnitTests/Helpers/Internals/Test_RuntimeHelpers.cs
@@ -1,0 +1,89 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if !NETCOREAPP3_1_OR_GREATER
+
+using System;
+using CommunityToolkit.HighPerformance.Helpers.Internals;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+#pragma warning disable CS0649
+
+namespace CommunityToolkit.HighPerformance.UnitTests.Helpers.Internals;
+
+[TestClass]
+public class Test_RuntimeHelpers
+{
+    [TestMethod]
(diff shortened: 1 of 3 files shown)


Arlodotexe (approved):

tests/CommunityToolkit.HighPerformance.UnitTests/Helpers/Internals/Test_RuntimeHelpers.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    private struct UnmanagedStruct2
+    {
+        public int A;
+        public UnmanagedStruct1 B;
+        public UnmanagedStruct1? C;
+        public Guid? D;
+    }
+
+    private struct GenericStruct<T>
```
XAML-Knight, line 79: I tested an extra assertion, and this passed, as well: 

`Assert.IsFalse(RuntimeHelpers.IsReferenceOrContainsReferences<GenericStruct<UnmanagedStruct1>>());`

XAML-Knight (approved):

tests/CommunityToolkit.HighPerformance.UnitTests/Helpers/Internals/Test_RuntimeHelpers.cs
Sergio0694: Added this test as well for good measure, why not! 😄

PR #69, opened 2021-12-17
Remove [Pure] attribute usages

Sergio0694:
This PR removes all `[Pure]` attribute usages from all .NET Community Toolkit libraries.
The underlying APIs are no longer used in .NET in general, so there's no point using these attributes.
Removing them also reduces the binary size a bit, so that's a small win as well.


Arlodotexe (approved):
Did a find-all in vscode, looks like all usages were removed 👍

XAML-Knight (approved):

PR #73, opened 2021-12-20
Add IsCurrency to StringExtensions

### LINKED ISSUE #72: Add IsCurrency to StringExtensions.
Continuing from [Currency validation support #4430](https://github.com/CommunityToolkit/WindowsCommunityToolkit/pull/4430#issuecomment-997855061).

We'd like to add an `IsCurrency` extension method to [StringExtensions](https://github.com/CommunityToolkit/dotnet/blob/main/CommunityToolkit.Common/Extensions/StringExtensions.cs) that bundles an easy and obvious check for if a string is a valid currency, similar to [IsDecimal](https://github.com/CommunityToolkit/dotnet/blob/main/CommunityToolkit.Common/Extensions/StringExtensions.cs#L66).

This would be implemented like so
```cs

public static bool IsCurrency([NotNullWhen(true)] this string? str)
{
    return decimal.TryParse(str, NumberStyles.Currency, out _);
}
```

Arlodotexe:
Closes #72.

Adds an `IsCurrency` extension method to StringExtensions that bundles an easy and obvious check for if a string is a valid currency, similar to `IsDecimal`.

### DIFF
--- a/CommunityToolkit.Common/Extensions/StringExtensions.cs
+++ b/CommunityToolkit.Common/Extensions/StringExtensions.cs
@@ -69,6 +69,17 @@ public static class StringExtensions
     }
 
     /// <summary>
+    /// Determines whether a string is a valid as a currency.
+    /// </summary>
+    /// <param name="str">The string to test.</param>
+    /// <param name="culture">The culture to check against. If left null, <see cref="CultureInfo.CurrentCulture"/> is used.</param>
+    /// <returns><c>true</c> for a valid currency; otherwise, <c>false</c>.</returns>
+    public static bool IsCurrency([NotNullWhen(true)] this string? str, CultureInfo? culture = null)
+    {
+        return decimal.TryParse(str, NumberStyles.Currency, culture ?? CultureInfo.CurrentCulture, out _);
+    }
+
+    /// <summary>
     /// Determines whether a string is a valid integer.
     /// </summary>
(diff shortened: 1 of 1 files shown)


Nirmal4G (approved):

Arlodotexe:
@Sergio0694 Whenever you're ready on this

faiteanu:
Hi, I'd like to draw the attention to several issues with this PR and would request to consider before merging it.

Does the name `IsCurrency` follow the .NET naming guidelines, specifically regarding misleading names? I would expect 
```csharp
"USD".IsCurrency == true;
"US Dollar".IsCurrency == true;
"MX".IsCurrency == false;
```
Your use of the word "currency" sounds non-standard. Maybe `IsCurrencyWithAmount`.

What is the global / world-wide use case, apart from your specific local use? IMO dealing with currency amounts without regarding the actual currency is a source for disaster. 
General tutorials give formatting examples like:
```csharp
// Default US Currency Format
string.Format(new CultureInfo("en-US"), "{0:C}", 1456.12155)      // $1,456.12
// Currency format for Mexico
string.Format(new CultureInfo("es-MX"), "{0:C}", 1456.12155)      // 1.456,12 Mex$
```
Please *never ever* write such code in a public application! If a price is 1,456.12 US Dollars, a customer cannot just pay 1.456,12 Mexican Pesos instead, just because he changed the language in his browser or operating system!

PR #87, opened 2022-01-13
Calculate the current C# compiler version

jmarolf:
This change allows us to calculate the current C# compiler version in a way that will work on all platforms and only relies on the `Microsoft.CSharp.CurrentVersion.targets` target (specifically [this line](https://github.com/dotnet/msbuild/blob/main/src/Tasks/Microsoft.CSharp.CurrentVersion.targets#L315)) being imported. `Microsoft.CSharp.CurrentVersion.targets` is the core msbuild logic for finding the C# compiler and is used universally so this fix should work everywhere there is a C# compiler to be found.


Sergio0694 (approved):
Tested on UWP (non SDK-style projects) and works great!
Thank you @jmarolf! 🎉

PR #90, opened 2022-01-13
Update base version number to reflect upcoming version

michael-hawker:
We should have our latest nightly builds show-up as the latest versions, so once we figure out a version number, we should update our version.json to match the upcoming release.


Sergio0694 (approved):
:shipit:

PR #91, opened 2022-01-13
Update .targets for MVVM Toolkit to fix .wapproj issues

Sergio0694:
This PR updates the .targets file for the MVVM Toolkit to resolve issues on .wapproj projects.


jmarolf (commented):
looks good!

PR #96, opened 2022-01-19
Centrally manage dependencies and its versions

Nirmal4G:
### Info

Fixes CommunityToolkit/WindowsCommunityToolkit#4180
Extracted from CommunityToolkit/WindowsCommunityToolkit#4181

### Changes

- Use Static Graph restore when applicable.
- Manage Packages Centrally using NuGet's CPVM feature.

Know more about NuGet's [**C**entral **P**ackage **V**ersion **M**anagement (CPVM)](https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions) feature.

Currently, NuGet package references are spread across project files. It is difficult to track versions and upgrade them properly. This patch enables NuGet's upcoming feature to manage package versions centrally using `Directory.Packages.props` with the following item spec to store a package version.

```xml
  <ItemGroup>
    <PackageVersion Include="Vendor.Awesome.Package" Version="11.0.0" />
  </ItemGroup>
```

### Notes

 - **Rebase** merge if possible.
 - Wait for my refactor PR to be merged!
 - When merging, please update the commit title to PR title instead of the default `Merge pull request #xxxx from repo/branch`, and commit message to either PR message or messages of individual commits. The `auto-merge` option does this by default.


Sergio0694 (changes requested):
I know this is work in progress, but please don't have these changes based on top of the other branch, let's keep PRs smaller. I'm saying this because I can't say when/if that base PR will get merged, so I don't think it's worth it to block this until that one is resolved, as it's unrelated 🙂
Thanks!

Nirmal4G:
There's a change in how we pack the SourceGen project into the parent project in this patch. I hope to create a new PR for that change but testing here before I do that. That change was done in response to this feature.

I don't want to update the workaround just to unblock the PR. This patch can wait as it needs support from VS IDE. As far as I know, v17.1 is the target to enable central package versioning in the NuGet PM UI.

Nirmal4G:
Please merge this only after merging #85 and #220 as it depends on some changes from those PRs. All the 3 PRs are correlated since the changes cascade into the other.

build/Community.Toolkit.Common.props
```diff
@@ -32,4 +32,19 @@
     <ContinuousIntegrationBuild>$(TF_BUILD)</ContinuousIntegrationBuild>
   </PropertyGroup>
 
-</Project>
+  <PropertyGroup>
+    <!-- Common Outputs -->
+    <BuildDir>~build\</BuildDir>
+    <PublishDir>~publish\</PublishDir>
```
Nirmal4G, line 38: Can we keep the build outputs in the project directory or move them to repository directory? If we prefer a common output path, we can move them to repository root.

```suggestion
    <!-- Common Outputs -->
    <BuildDir>$(RepositoryDirectory)~build\$(MSBuildProjectName)</BuildDir>
    <PublishDir>$(RepositoryDirectory)~publish\$(MSBuildProjectName)</PublishDir>
```

azure-pipelines.yml
```diff
@@ -9,27 +9,17 @@ pr:
...
   timeoutInMinutes: 60
   steps:
 
-  # Install NuGet
-  - task: NuGetToolInstaller@0
-    displayName: Install NuGet 6.0
-    inputs:
-      versionSpec: 6.0.0
-
-  # Install NerdBank GitVersioning
```
InKahootz, line 25: While the comment is wrong here, the command updates NBGV to the latest version if the pool image doesn't have it. So why remove it?

azure-pipelines.yml
```diff
@@ -38,35 +28,41 @@ jobs:
   - pwsh: build/Update-Headers.ps1 -Verify
     displayName: Verify headers
 
+  # Restore solution
+  - script: dotnet restore -p:Configuration=$(BuildConfiguration)
+    displayName: Restore solution
+
   # Build solution
-  - script: dotnet build -c Release
+  - script: dotnet build --no-restore -c $(BuildConfiguration)
```
InKahootz, line 36: `build` implicitly restores. Does the static graph restore need an explicit restore?

InKahootz (commented):
The PR is doing more than it states. Even if you leave off the first three commits which are part of your refactor. You're tackling two different issues

1. Building to a common path (a large portion of this PR)
2. Moving to centrally managed package versions

I think you should get feedback on 1. before implementing it as changing the build path to something you want is subjective. What's wrong with the current system?  

I've seen a lot of Microsoft repos have a common output that falls under `artifacts/`. Pretty much anything using the Arcade SDK.  I've never seen a convention of outputting to `~build` and `~packages` (I searched across GitHub and only found your personal projects using it).  

`artifacts/bin`, `artifacts/obj`, `artifacts/packages`, etc..

azure-pipelines.yml
Nirmal4G: Instead of cramming everything into a single step, I'm separating it, so that it'll be easy to look into the logs in CI.

azure-pipelines.yml
Nirmal4G: The pool does have the latest version and I read that the pool image gets updated according to Microsoft's Modern Lifecycle policy and .NET's Current channel.

Just another un-necessary step here.

Nirmal4G:
Initially, I wanted just two PRs, one for refactoring and another for central versioning but MVVM Changes are important enough to warrant an issue and PR of their own. But I can create a separate PR for common paths, if needed.

We can do `artifacts` if that's preferred strongly but having a `build`, `publish` and `packages` top-level directories makes sense, IMEX. And the tilde `~` symbol literally makes it top-level in the directory listing. That's why I choose this pattern.

<details>
<summary><h4>My thoughts… tl;dr M$FT is lazy!</h4></summary>

I experimented with Arcade SDK and found it lacking in so many ways, especially the defaults could've been better, IMHO! Explaining all the reasons why current defaults of .NET SDK is bad for large scale projects, would either need a blog series or a presentation talk!! 😂 Yeah, it's that bad!!! 🤦

And before you go on about how Microsoft uses Arcade for entire .NET Foundation projects, let me tell you, Arcade is simple defaults, low-config, too-many things, highly opinionated and tuned specific to only Microsoft managed .NET Foundation projects. It's goal is to take what M$FT repos would need and put them in a shared package so that everyone who needs it uses it without maintaining their own solution. It's not for general public.

_I prefer M$FT products over others anyday but not a fan! I used to in the days of XP, Vista and 7 but not now. They have become lazy!_

One thing you have to know about M$FT is that they do relatively minimal enough to get the project going since it requires a low maintainance cost, less work. It's how most corporates work and M$FT does it even less than others. They don't plan far ahead which is why most of their initiatives fail after a point in time.

It's like this, with M$FT, the final things they produce, be it software or hardware, they don't scale well, since the requirements are limited even though the ideas and design are good. Also, Exceptions are not examples!!

</details>

PR #111, opened 2022-02-14
Fix AsyncRelayCommand with allowConcurrentExecutions == false not raising CanExecuteChanged

### LINKED ISSUE #108: AsyncRelayCommand{T} with allowConcurrentExecutions==false doesn't Invoke CanExecuteChanged on execution
allowConcurrentExecutions alters the executability of AsyncRelayCommand{T}, however CanExecuteChanged isn't invoked, so any visual effects are not updated.

Sergio0694:
**Closes #108**

This PR fixes `AsyncRelayCommand` and `AsyncRelayCommand<T>` not raising `CanExecuteChanged` when invoked if `allowConcurrentExecutions` is set to `false` (which they should, given the command wouldn't be executable at that point).

### DIFF
--- a/tests/CommunityToolkit.Mvvm.UnitTests/Test_AsyncRelayCommand{T}.cs
+++ b/tests/CommunityToolkit.Mvvm.UnitTests/Test_AsyncRelayCommand{T}.cs
@@ -224,7 +224,7 @@ public class Test_AsyncRelayCommandOfT
         Assert.IsTrue(success);
     }
 
-    public void Test_AsyncRelayCommand_ExecuteDoesNotRaisesCanExecuteChanged()
+    public void Test_AsyncRelayCommand_ExecuteDoesNotRaiseCanExecuteChanged()
     {
         TaskCompletionSource<object?> tcs = new();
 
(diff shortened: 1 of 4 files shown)


tests/CommunityToolkit.Mvvm.UnitTests/Test_AsyncRelayCommand{T}.cs
```diff
@@ -223,4 +223,95 @@ async void TestCallback(Action throwAction, Action completeAction)
...
+        Assert.AreSame(command, args.Sender);
+        Assert.AreSame(EventArgs.Empty, args.Args);
+
+        Assert.IsFalse(command.CanExecute(""));
+
+        tcs.SetResult(null);
+    }
+
+    [TestMethod]
+    public void Test_AsyncRelayCommand_ExecuteDoesNotRaisesCanExecuteChanged_WithCancellation()
```
Arlodotexe, line 273: Minor typo, "Raises" -> "Raise"

Arlodotexe (approved):

tests/CommunityToolkit.Mvvm.UnitTests/Test_AsyncRelayCommand{T}.cs
Sergio0694: Whoops! Fixed in https://github.com/CommunityToolkit/dotnet/pull/111/commits/105e31ac8713124bd55351cf586c88bb9627eb18, thanks!

Issue #22, opened 2021-05-31
[Feature] AsyncRelayCommand.Execute should catch and raise an Exception if possible


juwens:
It easily happens that you use the AsyncRelayCommand somewhere as ICommand, if i call the ICommand.Execute i get no Exception, even if the async Method throws an Exception.

I know it's nearly impossible (only via AsyncEx.Run), to synchronous call an async method wait for it and raise the Exception in sync context.

### Alternative:

AsyncRelayCommand should have a public static Event where i can register an EventHandler for these kind of Exceptions.

So Execute would look like this:

```
public async void Execute()
{
    try
    {
        await ExecuteAsync()
    }
    catch(Exception e)
    {
        ExceptionInSyncExecute?.Invoke(this, new UnhandledExceptionEventArgs(e));
    }
}
```

timdinhdotcom:
We are using IAsyncCommand from Xamarin Community Toolkit, which has a parameter for `onException` action. It basically call the action whenever an exception occurred, which is very handy.

We are hoping to use `CommunityToolkit.Mvvm.Input.ICommandAttribute` instead to reduce our coding effort. Would be nice if it also support onException action too.

Suggest usage would be 
```
[ICommand(onException = SomeAction)]
private async Task MyCommand() {
    // code
}
```

Sergio0694:
Closing this, as exceptions can now be configured to be rethrown, via the new option added in the PR linked above.

PR #113, opened 2022-02-15
Propagate task exceptions to TaskScheduler.UnobservedTaskException

Sergio0694:
**Contributes to #22**

This PR adds a new internal `__TaskExtensions.GetAwaitableWithoutEndValidation` extensions (note: needs to be public as source generated code is using it too) that allows us to await a task without throwing exceptions. This has two main advantages:

- The code is now much more efficient, as we're no longer throwing/catching it, just ignoring it.
- Exceptions will now no longer be marked as observed, so they'll bubble up to `TaskScheduler.UnobservedTaskException`.

With this change, the MVVM Toolkit will essentially no longer interfere with task exceptions being observed, so users will be able to just monitor and log them through `TaskScheduler.UnobservedTaskException` as expected, unless manually observed.

### DIFF
--- a/tests/CommunityToolkit.Mvvm.UnitTests/Helpers/TaskSchedulerTestHelper.cs
+++ b/tests/CommunityToolkit.Mvvm.UnitTests/Helpers/TaskSchedulerTestHelper.cs
@@ -61,7 +61,12 @@ internal static class TaskSchedulerTestHelper
             // Await for the continuation to actually run
             _ = await tcs.Task;
 
-            // Some additional time to ensure the exception is propagated
+            // Wait for some additional time to ensure the exception is propagated. This is a bit counterintuitive, but the delay is
+            // not actually for the event to be raised, but to ensure the task that is throwing has had time to be scheduled and fail.
+            // The event is raised only when the exception wrapper inside that task is collected and its finalizer has run (that's where
+            // the logic to raise the event is executed), which is why we're then calling GC.Collect() and GC.WaitForPendingFinalizers().
+            // That is, we can't use a task completion source from that event, because that event is only guaranteed to actually be raised
+            // when the finalizer for that task run, which is why we're calling the GC after the delay here.
             await Task.Delay(200);
 
             GC.Collect();
--- a/tests/CommunityToolkit.Mvvm.UnitTests/Helpers/TaskSchedulerTestHelper.cs
+++ b/tests/CommunityToolkit.Mvvm.UnitTests/Helpers/TaskSchedulerTestHelper.cs
@@ -1,0 +1,92 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace CommunityToolkit.Mvvm.UnitTests.Helpers;
+
+/// <summary>
+/// A helper class to validate scenarios related to <see cref="TaskScheduler"/>.
+/// </summary>
+internal static class TaskSchedulerTestHelper
+{
+    /// <summary>
+    /// A custom <see cref="Delegate"/> for callbacks to <see cref="IsExceptionBubbledUpToUnobservedTaskExceptionAsync(TestCallback)"/>.
+    /// </summary>
+    /// <param name="throwAction">An <see cref="Action"/> instance that throws a test exception to track.</param>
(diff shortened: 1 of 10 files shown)


tests/CommunityToolkit.Mvvm.UnitTests/Helpers/TaskSchedulerTestHelper.cs
```diff
@@ -0,0 +1,92 @@
...
+            // to be a separate method to ensure the returned task isn't kept alive for longer.
+            callback(
+                () => throw new TestException(guid),
+                () => tcs.SetResult(null));
+
+            // Await for the continuation to actually run
+            _ = await tcs.Task;
+
+            // Some additional time to ensure the exception is propagated
+            await Task.Delay(200);
```
michael-hawker, line 65: You could use another TaskCompletionSource here and set the result in your function rather than using the bool `exceptionFound` directly. That way you ensure that if that code is going to run it's going to run. Main thing would be having a timeout for if the exception isn't thrown by the test then it doesn't just stall.

tests/CommunityToolkit.Mvvm.UnitTests/Test_AsyncRelayCommand.cs
```diff
@@ -225,4 +226,30 @@ public async Task Test_AsyncRelayCommand_WithConcurrencyControl()
 
         Assert.IsFalse(command.IsRunning);
     }
+
+    [TestMethod]
+    public async Task Test_AsyncRelayCommand_ThrowingTaskBubblesToUnobservedTaskException()
```
michael-hawker, line 231: Should an async call calling another async call which fails behave the same way? Do we want to have a test case for a more complex setup like that?

tests/CommunityToolkit.Mvvm.UnitTests/Helpers/TaskSchedulerTestHelper.cs
Sergio0694: So, this is a bit counterintuitive, but the delay is not actually for the event to be raised, but to ensure the task that is throwing has had time to be scheduled and fail. The event is raised only when the exception wrapper inside that task is collected and its finalizer has run (that's where the event is raised from), which is why we're then calling `GC.Collect()` and `GC.WaitForPendingFinalizers()`. That is, we can't use a task completion source from that event, because that event is only guaranteed to actually run when the finalizer for that task run, which is why we're calling the GC _after_ the delay here. Does that make sense? 😅

tests/CommunityToolkit.Mvvm.UnitTests/Test_AsyncRelayCommand.cs
Sergio0694: It should work the same way, yeah, the method would just throw when awaiting it, so the task would still end up being falted regardless, just with a different stack trace and exception (but we just care about some general failure) 🙂

michael-hawker:
@Sergio0694 want to add a comment to https://github.com/dotnet/csharplang/issues/5529 which outlines the Toolkit scenario their as well?

tests/CommunityToolkit.Mvvm.UnitTests/Helpers/TaskSchedulerTestHelper.cs
Arlodotexe: This extra insight would do good as a code comment here. Imagine having to refactor this without knowing that haha

Sergio0694: Yup, good idea! Done in 2ca1ac705571760127778f366a8a51613ce1fb43 👍

Arlodotexe (approved):

virzak:
Does this change the recommended way of handling errors occurring inside `AsyncRelayCommand`?

Sergio0694:
Hey @virzak, that's really up to you. What this change does is it stops commands from just completely swallowing exceptions, which made it impossible to easily log them through a handler listening to `UnobservedTaskException`. If that's enough for you, then sure. If instead you wanted some way to immediately detect an exception as soon as it occurs (the event is only raised when the GC kicks in, so it may take a bit), then you'd have to manually wire this up yourself by subscribing to `PropertyChanged`. Overall it depends on what you need in particular 🙂

virzak:
Cool, thanks, I was noticing it was firing after a while. Unfortunately still need to subscribe to `PropertyChanged` and can't use source generators.

PR #121, opened 2022-02-22
Fix analyzer build warnings/errors

Sergio0694:
This PR works around two analyzer issues that are causing build warnings and failures:

- RS2003 (https://github.com/dotnet/roslyn-analyzers/issues/5828)
- AD0001 (dotnet/linker#2642)

### DIFF
--- a/CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.csproj
+++ b/CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.csproj
@@ -4,11 +4,6 @@
     <TargetFrameworks>netstandard2.0;netstandard2.1;net6.0</TargetFrameworks>
   </PropertyGroup>
 
-  <!-- Workaround for .NET 6.0.x regression (see https://github.com/dotnet/linker/issues/2642) -->
-  <PropertyGroup>
-    <NoWarn>AD0001</NoWarn>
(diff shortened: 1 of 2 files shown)


agocke:
Confirmed the problem. This was fixed in main as part of my bootstrap addition, but that wasn't ported back to 6.0.2xx: https://github.com/dotnet/linker/pull/2540/files

CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.csproj
```diff
@@ -4,6 +4,11 @@
     <TargetFrameworks>netstandard2.0;netstandard2.1;net6.0</TargetFrameworks>
   </PropertyGroup>
 
+  <!-- Workaround for .NET 6.0.x regression (see https://github.com/dotnet/linker/issues/2642) -->
+  <PropertyGroup>
+    <NoWarn>AD0001</NoWarn>
```
agocke, line 9: I think you could narrow this suppression by suppressing the analyzer warning itself, i.e. IL2026.

Sergio0694: Thank you for the suggestion! Done in bc6f26f118094fc26a87814e99c6fdccdb700638 😊

PR #128, opened 2022-03-02
Add cancel command support

### LINKED ISSUE #118: APIs to create cancellation commands for AsyncRelayCommand types
### Overview and motivation

The MVVM Toolkit has the `AsyncRelayCommand` and `AsyncRelayCommand<T>` types, which make it easy to work with asynchronous operations through the usual `ICommand` pattern, while also adding support for cancellation, progress report, etc. They have one limitation though, which is that it's not easy to define a command to _cancel_ another command. Suppose you had this:

```csharp
partial class MyViewModel : ObservableObject
{
    [ICommand(AllowConcurrentExecutions = false)]
    private async Task DoStuffAsync(CancellationToken token)
    {
    }
}
```
```xml
<Button Content="Start" Command="{x:Bind ViewModel.DoStuffCommand}"/>
```

This works fine. But now suppose you wanted to also have a button to _cancel_ that operation. You'd need this:

```xml
<Button Content="Start" Command="{x:Bind ViewModel.DoStuffCommand}"/>
<Button
    Content="Cancel"
    Click="{x:Bind ViewModel.DoStuffCommand.Cancel}"
    IsEnabled="{x:Bind ViewModel.DoStuffCommand.CanBeCanceled}"
```

This works, but it's (1) not portable (eg. doesn't work if the framework doesn't allow binding to methods), and (2) it doesn't follow the same usual command pattern: we're binding an event directly to a method in the viewmodel. It would be nice to have a way to create a command to handle this, but this requires a rather clunky setup today, as you need to create a command wrapping the other one, and also subscribe to `PropertyChanged` on the first command, monitor for `CanBeCanceled` changing and then update the second command from there. It's not really user friendly and also quite error prone, plus it could be optimized if supported internally.

### API proposal

We could add an extension like this:

```csharp
namespace CommunityToolkit.Mvvm.Input;

public static class IAsyncRelayCommandExtensions
{
    public static ICommand CreateCancelCommand(this IAsyncRelayCommand command);
}
```

This could then be used as follows:

```csharp
class MyViewModel : ObservableObject
{
    public MyViewModel()
    {
        DoStuffCommand = new AsyncRelayCommand(DoStuffAsync);
        CancelDoStuffCommand = DoStuffCommand.CreateCancelCommand();
    }

    public IAsyncRelayCommand DoStuffCommand { get; }
    public ICommand CancelDoStuffCommand { get; }
    
    private async Task DoStuffAsync(CancellationToken token)
    {
    }
}
```
```xml
<Button Content="Start" Command="{x:Bind ViewModel.DoStuffCommand}"/>
<Button Content="Cancel" Command="{x:Bind ViewModel.CancelDoStuffCommand}"/>
```

We could then also add some way to automatically create a cancel command through source generators, like:

```csharp
partial class MyViewModel : ObservableObject
{
    [ICommand(AllowConcurrentExecutions = false, IncludeCancelCommand = true)]
    private async Task DoStuffAsync(CancellationToken token)
    {
    }
}
```

Which would generate the pair of command and cancel command.

Thoughts? Other API ideas or alternatives? 😄

cc. @michael-hawker @Arlodotexe

Sergio0694:
**Closes #118**

This PR adds the following changes:

```diff
 namespace CommunityToolkit.Mvvm.Input;

 public sealed class ICommandAttribute : Attribute
 {
     public string? CanExecute { get; init; }
     public bool AllowConcurrentExecutions { get; init; }
+    public bool IncludeCancelCommand { get; init; }
 }

+public static class IAsyncRelayCommandExtensions
+{
+    public static ICommand CreateCancelCommand(this IAsyncRelayCommand command);
+}
```


tests/CommunityToolkit.Mvvm.UnitTests/Test_ICommandAttribute.cs
```diff
@@ -567,4 +616,45 @@ public sealed class User
...
+            catch (OperationCanceledException e)
+            {
+                _ = Tcs1.TrySetResult(e);
+            }
+        }
+
+        [ICommand(IncludeCancelCommand = true)]
+        private async Task DoWorkWithParameterAsync(int number, CancellationToken token)
+        {
+            using CancellationTokenRegistration registration = token.Register(static state => ((TaskCompletionSource<object?>)state!).TrySetCanceled(), Tcs2);
```
Arlodotexe, line 646: Have you made sure this task fails successfully when a nullref happens on the suppressed nullable? I've seen some cases where the MSTest framework doesn't fail properly when an exception throws inside a task, sometimes waiting indefinitely, returning inconclusive, or just passing anyway.

Arlodotexe (approved):

tests/CommunityToolkit.Mvvm.UnitTests/Test_ICommandAttribute.cs
Sergio0694: That can never be null, because I'm passing that `Tcs2` instance as state, which is not nullable.
It's just that `CancellationToken.Register` doesn't know the state nullability propagates to the state callback.

PR #129, opened 2022-03-02
Dev/cancel commands

Micke3rd:
This PR removes redundant code elements from *RelayCommand classes.
The PR is towards the "Dev/cancel commands" branch, because these classes there touched at last, although the redundant elements are older.

@Sergio0694 thank you for taking a look at the CancelCommand topic.


Sergio0694:
Hi @Micke3rd, thank you for your interest in the MVVM Toolkit. This PR includes an API difference, and as such it should first be discussed in a proposal (a new issue), and then eventually done in a PR if the proposal is approved. In this specific case, I can also tell you that we don't really have plans to introduce a new `CommandBase` type to the MVVM Toolkit.
Thanks again! 🙂

PR #135, opened 2022-03-06
Add [DynamicallyAccessedMembers] annotations

Sergio0694:
This PR adds `[DynamicallyAccessedMembers]` to the generated methods for `ObservableValidator` and `IRecipient<TMessage>`. Since the type is only available on .NET 6 and above, the additional attributes is only generated when that target is being used.


CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableValidatorValidateAllPropertiesGenerator.Execute.cs
```diff
@@ -108,20 +108,56 @@ public static RecipientInfo GetInfo(INamedTypeSymbol typeSymbol, ImmutableArray<
         /// <summary>
         /// Gets the head <see cref="CompilationUnitSyntax"/> instance.
         /// </summary>
+        /// <param name="isDynamicallyAccessedMembersAttributeAvailable">Indicates whether <c>[DynamicallyAccessedMembers]</c> should be generated.</param>
         /// <returns>The head <see cref="CompilationUnitSyntax"/> instance with the type attributes.</returns>
-        public static CompilationUnitSyntax GetSyntax()
+        public static CompilationUnitSyntax GetSyntax(bool isDynamicallyAccessedMembersAttributeAvailable)
         {
+            int numberOfAttributes = 5 + (isDynamicallyAccessedMembersAttributeAvailable ? 1 : 0);
```
Arlodotexe, line 115: What is this 5 determined by? Is it possible to get this number programmatically, rather than hardcoding?
If not, this should probably have a comment on it.

tests/CommunityToolkit.Mvvm.UnitTests/Test_ObservableRecipient.cs
```diff
@@ -85,6 +92,23 @@ public void Test_ObservableRecipient_Broadcast(Type type)
         Assert.AreEqual(message.PropertyName, nameof(SomeRecipient<int>.Data));
     }
 
+    [TestMethod]
+    public void Test_IRecipient_VerifyTrimmingAnnotation()
```
Arlodotexe, line 96: This test doesn't appear to be tied to `IRecipient` and is a 1:1 copy of `Test_ObservableValidator_VerifyTrimmingAnnotation`.

Arlodotexe (changes requested):

tests/CommunityToolkit.Mvvm.UnitTests/Test_ObservableRecipient.cs
Sergio0694: Oh whoops, forgot to update this one. Fixed in https://github.com/CommunityToolkit/dotnet/pull/135/commits/5fb9f7039cef0621684b3580726d6e5c5c9891aa, thanks! 😄

CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableValidatorValidateAllPropertiesGenerator.Execute.cs
Arlodotexe: Yeah I see that, just wondering if hardcoding is a good idea

tests/CommunityToolkit.Mvvm.UnitTests/Test_ObservableRecipient.cs
Arlodotexe: Thanks, Sergio. Though this test still doesn't seem to be related to `IRecipient` specifically. Maybe rename the method to clear up confusion?

```suggestion
    public void Test_IMessenger_VerifyTrimmingAnnotation()
```

Arlodotexe (approved):

tests/CommunityToolkit.Mvvm.UnitTests/Test_ObservableRecipient.cs
Sergio0694: I think `IRecipient` is more appropriate because the generator specifically looks for that interface.
The fact you're using `IMessenger` is not really tied to the logic of the generator itself here.

PR #151, opened 2022-03-14
Move collection types to MVVM Toolkit

### LINKED ISSUE #138: Move grouped collections to MVVM Toolkit
All grouped observable collections should be moved from the Common package to the MVVM Toolkit.

Sergio0694:
**Closes #138**

This PR moves the collection types with no further changes, other than the namespace.
Follow up API changes will be done in a separate PR.


jamesmontemagno:
🎉🎉🎉🎉

jamesmontemagno:
Will this have things like AddRange, remove range, etc? Similar to https://github.com/jamesmontemagno/mvvm-helpers/blob/master/MvvmHelpers/ObservableRangeCollection.cs

Sergio0694:
Hey @jamesmontemagno, that's actually something that I specifically don't want to add. We've talked about this with @michael-hawker and @brminnick (who also had the same question, as they currently have similar helpers in the MAUI Community Toolkit, which are being removed as they standardize towards the MVVM Toolkit), and my position was that we should instead try to help make https://github.com/dotnet/runtime/pull/65101 happen, and then just use that. Currently the PR is in the works, we mostly just need to talk to the WinUI 3 team to see if they can support it too (not being open source we can't just contribute to it like with WPF), and confirm that MAUI will support that too (maybe you and Brandon can help clarify this?). With that, it'll be a much better end result for consumers, as the API having proper support both has much better performance, as well as offering better UI animations (the UI can only animate the replaces items instead of resetting the whole collection). As we said before, the goal here is to put together the best possible API set for consumers in the long term, and we feel like this is the right direction to achieve that 🙂

Hope this helps!

michael-hawker:
Anything new or added here? This is just a namespace/package change, right? Curious why they're not all showing as moves.

Sergio0694:
@michael-hawker Yes there's a bunch of changes I want to discuss in today's API review 🙂

michael-hawker:
Capturing notes from discussion today:
1. Separate this into two PRs, one for the file/namespace change only.
2. New PR for the functional changes.

That'll make things easier to review. :)

For the new PR:
- We also discussed cleaning up the interfaces to align better to the BCL and provide additional value with other existing interfaces like `ILookup`.
- It also was discussed to evaluate and change the extension methods to not clash with LINQ methods and either rename or align with better docs/samples in the future.

jamesmontemagno:
Ahhhh i gotcha now, yeah that makes sense looking at the docs what this is. Still great ;)

Sergio0694:
YOLO-merging this one since it's just the move we discussed, so I can open the new PR with API changes 😄

PR #160, opened 2022-03-16
Fix nullability annotations in generated properties

### LINKED ISSUE #155: Generated code needs to be with the same nullability context as defined by the user
Using v8.0.0-preview2 NuGet package.

Have defined property of type `List<string?>` and decorated it with `[ObservableProperty]` attribute.

MvvmToolkit-User-Code

But the generated code is not nullable enabled, because of which getting warnings in other places.

MvvmToolkit-Generated-Code

Nullability context needs to be taken care in the source generator.

Sergio0694:
**Closes #155**

This PR fixes the generated properties to preserve the same nullability context as the source fields.

### DIFF
--- a/tests/CommunityToolkit.Mvvm.UnitTests/Test_ObservablePropertyAttribute.cs
+++ b/tests/CommunityToolkit.Mvvm.UnitTests/Test_ObservablePropertyAttribute.cs
@@ -509,11 +509,9 @@ public partial class Test_ObservablePropertyAttribute
         //Assert.AreEqual(NullabilityState.NotNull, rightInfo2.ReadState);
         //Assert.AreEqual(NullabilityState.NotNull, rightInfo2.WriteState);
 
-        // The commented out lines are to work around a weird behavior of the NullabilityInfo API there.
-        // Arguably we're pushing them a bit far here, but it's fine. Even with those cases commented out,
-        // the test is already more than enough, plus we can also double check the behavior by looking at
-        // the generated code. Thoe lines can be uncommented once the behavior is either clarified, or if
-        // it happens to be a bug which is then fixed in a future version of .NET, once we upgrade as well.
+        // The commented out lines are to work around a bug in the NullabilityInfo API in .NET 6.
+        // This has been fixed for .NET 7: https://github.com/dotnet/runtime/pull/63556. The test
+        // cases above can be uncommented when the .NET 7 target (or a more recent version) is added.
     }
 #endif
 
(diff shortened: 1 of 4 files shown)


tests/CommunityToolkit.Mvvm.UnitTests/Test_ObservablePropertyAttribute.cs
```diff
@@ -360,6 +361,105 @@ public void Test_OnPropertyChangingAndChangedPartialMethodWithAdditionalValidati
...
+        NullabilityInfo info = context.Create(typeof(NullableRepro).GetProperty(nameof(NullableRepro.NullableMess))!);
+
+        Assert.AreEqual(typeof(Foo<Foo<string?, int>.Bar<object?>?, StrongBox<Foo<int, string?>.Bar<object>?>?>), info.Type);
+        Assert.AreEqual(NullabilityState.Nullable, info.ReadState);
+        Assert.AreEqual(NullabilityState.Nullable, info.WriteState);
+        Assert.AreEqual(2, info.GenericTypeArguments.Length);
+
+        NullabilityInfo leftInfo = info.GenericTypeArguments[0];
+
+        Assert.AreEqual(typeof(Foo<string?, int>.Bar<object?>), leftInfo.Type);
```
Arlodotexe, line 400: Since this is such a large method, it might be helpful to add a few comments to visualize which part of the generic is being tested. It's not immediately obvious with the current variable names.

tests/CommunityToolkit.Mvvm.UnitTests/Test_ObservablePropertyAttribute.cs
```diff
@@ -360,6 +361,105 @@ public void Test_OnPropertyChangingAndChangedPartialMethodWithAdditionalValidati
...
+        Assert.AreEqual(NullabilityState.Nullable, rightInfo1.ReadState);
+        Assert.AreEqual(NullabilityState.Nullable, rightInfo1.WriteState);
+
+        NullabilityInfo rightInfo2 = rightInnerInfo.GenericTypeArguments[2];
+
+        Assert.AreEqual(typeof(object), rightInfo2.Type);
+        //Assert.AreEqual(NullabilityState.NotNull, rightInfo2.ReadState);
+        //Assert.AreEqual(NullabilityState.NotNull, rightInfo2.WriteState);
+
+        // The commented out lines are to work around a weird behavior of the NullabilityInfo API there.
```
Arlodotexe, line 455: I think you mentioned offline that there was an issue filed with the Rosyln team on this? Could you link it in the comment?

Arlodotexe (approved):

tests/CommunityToolkit.Mvvm.UnitTests/Test_ObservablePropertyAttribute.cs
Sergio0694: Done in https://github.com/CommunityToolkit/dotnet/pull/160/commits/393a2232296dd90cb3cc96977f1728a88e52a8a4 👍

PR #161, opened 2022-03-16
Allow Task<T>-returns on methods with [ICommand]

### LINKED ISSUE #159: [ICommand] should support Task<T>-returning methods
Currently, `[ICommand]` will show an error if used on a method returning a `Task<T>` object, but that shouldn't be the case since such a method would be valid when used as argument for the constructors for an async relay command. The generator should reflect this and allow methods with this signature too.

Sergio0694:
**Closes #159**

This PR allows `[ICommand]` to also be used on methods returning `Task<T>`.


kronic:
Add support ValueTask and ValueTask<T>

Arlodotexe (approved):

Sergio0694:
@kronic That's outside of the scope of this PR, not planned, and doesn't have an approved proposal 🙂

PR #177, opened 2022-03-24
Added Check APIs

### LINKED ISSUE #176: Check (a partner API to Guard)
### Overview

This is very similar to #144, except instead of being part of Guard, Check is its own API.

Check would contain an equivalent method for each Guard method that returns a simple true or false instead of throwing

### API breakdown

A null check would look like this

```cs
public static class Check
{
    /// <summary>
    /// Checks that the input value is <see langword="null"/>.
    /// </summary>
    /// <typeparam name="T">The type of reference value type being tested.</typeparam>
    /// <param name="value">The input value to test.</param>
    /// <returns>True if the value is null, False otherwise.</returns>
    public static bool IsNull<T>(T? value)
    {
        return value is null;
    }
}
```

Which is extremely similar to the null guard's

```cs
public static class Guard
{
    /// <summary>
    /// Asserts that the input value is <see langword="null"/>.
    /// </summary>
    /// <typeparam name="T">The type of reference value type being tested.</typeparam>
    /// <param name="value">The input value to test.</param>
    /// <param name="name">The name of the input parameter being tested.</param>
    /// <exception cref="ArgumentException">Thrown if <paramref name="value"/> is not <see langword="null"/>.</exception>
    public static void IsNull<T>(T? value, [CallerArgumentExpression("value")] string name = "")
    {
        if (value is null)
        {
            return;
        }

        ThrowHelper.ThrowArgumentExceptionForIsNull(value, name);
    }
}
```

In practice such a simple check would not be used alone, but having a check equivalent for every guard would provide consistency and the more elaborate checks are useful.

### Usage example

```cs
public struct UniformKernel
{
  private double _min;
  private double _max;
  
  public double Weight(double value)
  {
    return Check.IsBetween(value, _min, _max) ? 1 : 0;
  }
}
```

In more elaborate contexts it might just return a sentinel value early

### Breaking change?

No

### Alternatives

This is just candy. The alternative is to check these simple conditions where needed without a helper method.

### Additional context

_No response_

### Help us help you

Avid29:
**Closes #176**

## PR Checklist

- [x] New component
       (By making Guard use Check they are tested together)


michael-hawker:
Converted to draft as waiting on #176 issue to be approved. @Sergio0694 do you want a label for this? We should investigate if the bot understands linked issues/PRs or not, is the fabric bot running on this repo yet?

PR #181, opened 2022-03-26
Fix [ICommand] generator crashing with overloads, add diagnostics

### LINKED ISSUE #154: Duplicating a method with the ICommand attribute breaks source generator
Not sure if it's a bigger VS problem or not, but it's a bit disruptive.  Add the following to a view model:

```csharp
[ICommand]
private void DoSomething()
{
}
```

Now, highlight all three lines and press `Ctrl-D` (or copy and paste) at least 3 times.  The ICommandSourceGenerator breaks, the previously generated files disappear and even after adding a unique suffix to each method, it won't create any new files.  The same thing doesn't happen with:

```csharp
[ObservableProperty] int foo;
```

Sergio0694:
**Closes #154**

## PR Checklist


Sergio0694:
@Arlodotexe they should be handled, yeah, since I'm just enumerating all method symbols from the containing type symbol by name and then checking whether the symbols are the same, so the exact signature doesn't really matter 🙂

PR #188, opened 2022-03-30
WIP return valid values from Guard methods

### LINKED ISSUE #187: Guard methods should return values when valid
### Overview

Currently the `Guard` methods are all declared as `void`, meaning they can't be used inline. If the methods returned the value passed in, they could be used in assignments.

Right now, you have to write this:

```csharp
public class Person
{
    public Person(string name)
    {
        Guard.IsNotNullOrEmpty(name);
        Name = name;
    }
    public string Name { get; }
}
```

If the `Guard` method returned the value you could halve the number of lines in constructors like this.

### API breakdown

As an example, the `IsNotNullOrEmpty` method would look like this:

```csharp
public static string IsNotNullOrEmpty([NotNull] string? text, [CallerArgumentExpression("text")] string name = "")
{
    if (string.IsNullOrEmpty(text))
    {
        ThrowHelper.ThrowArgumentExceptionForIsNotNullOrEmpty(text, name);
    }

    return text!;
}
```

### Usage example

```csharp
public class Person
{
    public Person(string name)
    {
        Name = Guard.IsNotNullOrEmpty(name);
    }
    public string Name { get; }
}
```

### Breaking change?

No

### Alternatives

None?

### Additional context

_No response_

### Help us help you

markrendle:
**Closes #187**

Changes Guard methods to return the passed value when it is valid.

I've only made the changes in `Guard.cs` and `Guard.String.cs` so far as a starting point for discussion. Also haven't changed tests or updated XML doc comments.

## PR Checklist

- [ ] New component

## Other information

I don't *think* this would be a breaking change but I am open to being corrected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,9 @@ bld/
 # Visual Studio 2015 cache/options directory
 .vs/
 
+# Rider cache/options directory
+.idea/
+
 # MSTest test Results
 [Tt]est[Rr]esult*/
 [Bb]uild[Ll]og.*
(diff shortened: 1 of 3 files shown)


Sergio0694:
Closing this as the proposal has been rejected. For the future, as the message say in the PR template as well, don't open a PR before the related proposal has been approved, or the PR would just be rejected immediately by policy. Thank you! 🙂

PR #195, opened 2022-04-01
Match methods/checks added to the Guard.

mabroukmahdhi:
# New check Methods
## Object -> Match\<T>: 
Asserts that the predicate is satisfied.

```csharp
 public static void Match<T>(T value, Expression<Func<T, bool>> predicate, [CallerArgumentExpression("value")] string name = "")
```

## Object -> NotMatch\<T>: 
Asserts that the predicate is not satisfied.

```csharp
 public static void NotMatch<T>(T value, Expression<Func<T, bool>> predicate, [CallerArgumentExpression("value")] string name = "")
```

## String -> Match: 
Asserts that the regular expression is matched.

```csharp
public static void Match(string value, string regexPattern, [CallerArgumentExpression("value")] string name = "")
```

## String -> NotMatch: 
Asserts that the regular expression is not matched.

```csharp
public static void NotMatch(string value, string regexPattern, [CallerArgumentExpression("value")] string name = "")
```

### DIFF
--- a/CommunityToolkit.Diagnostics/Guard.String.cs
+++ b/CommunityToolkit.Diagnostics/Guard.String.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
 
 #pragma warning disable CS8777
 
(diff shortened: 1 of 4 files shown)


Sergio0694:
Closing this. As the PR template clearly states, PRs opened without an approve API proposal will not be accepted.
You're welcome to open a proper API proposal if you'd like, and then we can go from there 🙂

PR #199, opened 2022-04-05
Add EnsureCapacity to ArrayPoolExtensions

### LINKED ISSUE #191: Api Proposal: ArrayPool extension "EnsureCapacity"
### Overview

When working with ReadOnlySequences or converting strings to binary data (or vice versa) I've found it useful to use a pooled array as a temporary buffer when needed.

The buffer should be rented only when needed and sliced if it's too big, which is a different use case compared to the existing `Resize`-extension.
Doing null/length checks manually is not complex, but possibly a bit error prone and verbose.

I'm not overly attached to the `EnsureCapacity`-name, but it's the best I've come up with. The functionality is also loosely related to EnsureCapacity on List and Dictionary.

### API breakdown

Ensures that when the method returns, `array` is not null and is at least `capacity` in length. In case the array needs to be "resized", contents are not preserved.

Error handling for negative capacity can be left to the pool.

```csharp
namespace CommunityToolkit.HighPerformance;

public static class ArrayPoolExtensions
{
    public static void EnsureCapacity<T>(
        this ArrayPool<T> pool,
        [NotNull] ref T[]? array,
        int capacity,
        bool clearArray = false)
    {
        if (array is null)
        {
            array = pool.Rent(capacity);
        }
        else if (array.Length < capacity)
        {
            var newArray = pool.Rent(capacity);
            pool.Return(array, clearArray);
            array = newArray;
        }
    }
}
```

### Usage example

```csharp
byte[]? buffer = null;

try
{
    foreach (ReadOnlySequence<byte> sequence in SomeReader())
    {
        if (sequence.IsSingleSegment)
        {
            ParseSpan(sequence.FirstSpan);
        }
        else
        {
            ArrayPool<byte>.Shared.EnsureCapacity(ref buffer, (int)sequence.Length);
            sequence.CopyTo(buffer);
            ParseSpan(buffer.AsSpan(0, (int)sequence.Length));
        }
    }
}
finally
{
    if (buffer is not null)
        ArrayPool<byte>.Shared.Return(buffer);
}
```

### Breaking change?

No

### Alternatives

The method is relatively simple, possibly rarely used, and can be copy-pasted wherever it's needed; simplest alternative is to just not implement it.

I've implemented it a couple of times in projects already referencing the HighPerformance-package. I can live with it if the extra api surface isn't worth the cost.

I'm interested in hearing everyone's thoughts on this. 

### Additional context

_No response_

### Help us help you

ovska:
**Closes #191**

Implements an extension `EnsureCapacity` for `ArrayPool<T>` that ensures that an array is not null and has a minimum length. Documentation closely mirrors the existing `Resize`-method for consistency. The only difference to the original proposal is the range check for `capacity`, as requested.

~~Added `[DoesNotReturn]` "polyfill" for under NETSTANDARD2.1. This isn't critical and can be removed if deemed not to be worth it.~~

## PR Checklist

## Other information

Renamed weird/outdated test method names for existing `ArrayPoolExtensions.Resize`-tests so new tests fit in better.

### DIFF
--- a/CommunityToolkit.HighPerformance/Extensions/ArrayPoolExtensions.cs
+++ b/CommunityToolkit.HighPerformance/Extensions/ArrayPoolExtensions.cs
@@ -53,4 +53,45 @@ public static class ArrayPoolExtensions
 
         array = newArray;
     }
+
+    /// <summary>
+    /// Ensures that when the method returns <paramref name="array"/> is not null and is at least <paramref name="capacity"/> in length.
+    /// Contents of <paramref name="array"/> are not copied if a new array is rented.
+    /// </summary>
+    /// <typeparam name="T">The type of items into the target array given as input.</typeparam>
+    /// <param name="pool">The target <see cref="ArrayPool{T}"/> instance used to rent and/or return the array.</param>
+    /// <param name="array">The rented <typeparamref name="T"/> array to ensure capacity for, or <see langword="null"/> to rent a new array.</param>
+    /// <param name="capacity">The minimum length of <paramref name="array"/> when the method returns.</param>
+    /// <param name="clearArray">Indicates whether the contents of the array should be cleared if returned to the pool.</param>
+    /// <exception cref="ArgumentOutOfRangeException">Thrown when <paramref name="capacity"/> is less than 0.</exception>
+    /// <remarks>When this method returns, the caller must not use any references to the old array anymore.</remarks>
+    public static void EnsureCapacity<T>(this ArrayPool<T> pool, [NotNull] ref T[]? array, int capacity, bool clearArray = false)
+    {
+        if (capacity < 0)
+        {
+            ThrowArgumentOutOfRangeExceptionForNegativeArrayCapacity();
+        }
+
+        if (array is null)
+        {
+            array = pool.Rent(capacity);
+        }
+        else if (array.Length < capacity)
+        {
+            // Ensure rent succeeds before returning the original array to the pool
+            T[] newArray = pool.Rent(capacity);
+
+            pool.Return(array, clearArray);
+
+            array = newArray;
+        }
+    }
+
+    /// <summary>
+    /// Throws an <see cref="ArgumentOutOfRangeException"/> when the "capacity" parameter is negative.
+    /// </summary>
+    private static void ThrowArgumentOutOfRangeExceptionForNegativeArrayCapacity()
+    {
+        throw new ArgumentOutOfRangeException("capacity", "The array capacity must be a positive number.");
+    }
 }
--- a/tests/CommunityToolkit.HighPerformance.UnitTests/Extensions/Test_ArrayPoolExtensions.cs
+++ b/tests/CommunityToolkit.HighPerformance.UnitTests/Extensions/Test_ArrayPoolExtensions.cs
@@ -87,4 +87,73 @@ public class Test_ArrayPoolExtensions
         Assert.AreNotSame(array, backup);
         Assert.IsTrue(backup.AsSpan(0, 16).ToArray().All(i => i == 0));
     }
+
+    [TestMethod]
+    [ExpectedException(typeof(ArgumentOutOfRangeException))]
+    public void Test_ArrayPoolExtensions_EnsureCapacity_InvalidCapacity()
+    {
+        int[]? array = null;
+
+        ArrayPool<int>.Shared.EnsureCapacity(ref array, -1);
+    }
+
+    [TestMethod]
+    public void Test_ArrayPoolExtensions_EnsureCapacity_IdenticalCapacity()
+    {
+        int[]? array = ArrayPool<int>.Shared.Rent(10);
+        int[]? backup = array;
+
+        ArrayPool<int>.Shared.EnsureCapacity(ref array, 10);
+        Assert.AreSame(backup, array);
+        Assert.IsTrue(array.Length >= 10);
+    }
+
+    [TestMethod]
+    public void Test_ArrayPoolExtensions_EnsureCapacity_NewArray()
+    {
+        int[]? array = null;
+
+        ArrayPool<int>.Shared.EnsureCapacity(ref array, 7);
+
+        Assert.IsNotNull(array);
+        Assert.IsTrue(array.Length >= 7);
+        int[]? backup = array;
+
+        ArrayPool<int>.Shared.EnsureCapacity(ref array, 64);
+
+        Assert.AreNotSame(backup, array);
+        Assert.IsTrue(array.Length >= 64);
+    }
+
+    [TestMethod]
+    public void Test_ArrayPoolExtensions_EnsureCapacity_SufficientCapacity()
+    {
+        int[]? array = ArrayPool<int>.Shared.Rent(16);
+        int[]? backup = array;
+
+        ArrayPool<int>.Shared.EnsureCapacity(ref array, 8);
+        Assert.AreSame(backup, array);
+
+        ArrayPool<int>.Shared.EnsureCapacity(ref array, 16);
+        Assert.AreSame(backup, array);
+
+        ArrayPool<int>.Shared.EnsureCapacity(ref array, 0);
+        Assert.AreSame(backup, array);
+    }
+
+    [TestMethod]
+    public void Test_ArrayPoolExtensions_EnsureCapacity_ClearArray()
+    {
+        int[]? array = ArrayPool<int>.Shared.Rent(16);
+        int[]? backup = array;
+
+        array.AsSpan().Fill(7);
+        Assert.IsTrue(backup.All(i => i == 7));
+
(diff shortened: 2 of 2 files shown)


CommunityToolkit.HighPerformance/Extensions/ArrayPoolExtensions.cs
```diff
@@ -53,4 +53,41 @@ public static void Resize<T>(this ArrayPool<T> pool, [NotNull] ref T[]? array, i
 
         array = newArray;
     }
+
+    /// <summary>
+    /// Ensures that when the method returns <paramref name="array"/> is not null and is at least <paramref name="capacity"/> in length.
+    /// Contents of <paramref name="array"/> are not copied if a new array is rented.
+    /// </summary>
+    /// <typeparam name="T">Array item type</typeparam>
```
Sergio0694, line 61:
```suggestion
    /// <typeparam name="T">The type of items into the target array given as input.</typeparam>
```

CommunityToolkit.HighPerformance/Extensions/ArrayPoolExtensions.cs
```diff
@@ -53,4 +53,41 @@ public static void Resize<T>(this ArrayPool<T> pool, [NotNull] ref T[]? array, i
...
+    /// <param name="pool">The target <see cref="ArrayPool{T}"/> instance used to rent and/or return the array.</param>
+    /// <param name="array">The rented <typeparamref name="T"/> array to ensure capacity for, or <see langword="null"/> to rent a new array.</param>
+    /// <param name="capacity">Minimum length of <paramref name="array"/> when the method returns.</param>
```
Sergio0694, line 64:
```suggestion
    /// <param name="capacity">The minimum length of <paramref name="array"/> when the method returns.</param>
```

CommunityToolkit.HighPerformance/Extensions/ArrayPoolExtensions.cs
```diff
@@ -53,4 +53,41 @@ public static void Resize<T>(this ArrayPool<T> pool, [NotNull] ref T[]? array, i
...
+            ThrowNegativeCapacity();
+        }
+
+        if (array is null)
+        {
+            array = pool.Rent(capacity);
+        }
+        else if (array.Length < capacity)
+        {
+            // ensure rent succeeds before returning the original array
```
Sergio0694, line 81: Nit: comments should start in uppercase
```suggestion
            // Ensure rent succeeds before returning the original array
```

CommunityToolkit.HighPerformance/Extensions/ArrayPoolExtensions.cs
```diff
@@ -53,4 +53,41 @@ public static void Resize<T>(this ArrayPool<T> pool, [NotNull] ref T[]? array, i
...
+            T[] newArray = pool.Rent(capacity);
+            pool.Return(array, clearArray);
+            array = newArray;
```
Sergio0694, line 84: Code style: add a blank line between declarations, invocations and assignments.
```suggestion
            T[] newArray = pool.Rent(capacity);

            pool.Return(array, clearArray);

            array = newArray;
```

CommunityToolkit.HighPerformance/Extensions/ArrayPoolExtensions.cs
```diff
@@ -53,4 +53,41 @@ public static void Resize<T>(this ArrayPool<T> pool, [NotNull] ref T[]? array, i
...
+        }
+
+        [DoesNotReturn]
+        static void ThrowNegativeCapacity()
+        {
+            throw new ArgumentOutOfRangeException(nameof(capacity));
```
Sergio0694, line 90: Given this is a throw helper anyway, we might as well add a helpful error message as well:
```suggestion
            throw new ArgumentOutOfRangeException(nameof(capacity), "The array capacity must be a positive number.");
```

CommunityToolkit.HighPerformance/Extensions/ArrayPoolExtensions.cs
```diff
@@ -53,4 +53,41 @@ public static void Resize<T>(this ArrayPool<T> pool, [NotNull] ref T[]? array, i
...
+        else if (array.Length < capacity)
+        {
+            // ensure rent succeeds before returning the original array
+            T[] newArray = pool.Rent(capacity);
+            pool.Return(array, clearArray);
+            array = newArray;
+        }
+
+        [DoesNotReturn]
+        static void ThrowNegativeCapacity()
```
Sergio0694, line 88: Nit: use a more expressive method name:

```suggestion
        static void ThrowArgumentOutOfRangeExceptionForNegativeCapacity()
```

tests/CommunityToolkit.HighPerformance.UnitTests/Extensions/Test_ArrayPoolExtensions.cs
```diff
@@ -87,4 +87,62 @@ public void Test_ArrayExtensions_Clear()
...
+    [ExpectedException(typeof(ArgumentOutOfRangeException))]
+    public void Test_ArrayPoolExtensions_EnsureCapacity_InvalidSize()
+    {
+        int[]? array = null;
+
+        ArrayPool<int>.Shared.EnsureCapacity(ref array, -1);
+    }
+
+    [TestMethod]
+    public void Test_ArrayPoolExtensions_EnsureCapacity_NewArray()
```
Sergio0694, line 101: Could you also add a test with the input array being of exactly the same capacity as the new requested capacity?

CommunityToolkit.HighPerformance/Properties/Polyfills/DoesNotReturnAttribute.cs
```diff
@@ -0,0 +1,13 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if !NETSTANDARD2_1_OR_GREATER
+
+namespace System.Diagnostics.CodeAnalysis;
+
+/// <summary>Applied to a method that will never return under any circumstance.</summary>
+[AttributeUsage(AttributeTargets.Method, Inherited = false)]
+public sealed class DoesNotReturnAttribute : Attribute { }
```
Sergio0694, line 11: This cannot be public, at best it should be internal. But regardless, I don't think we should add this in this PR, given it's not really needed and all the other throw helpers in the paclage are not using this attribute either. If we really wanted to do this, it should be a separate follow up PR just doing that 🙂

Sergio0694 (changes requested):
Looking good! Left some minor nits 😊

CommunityToolkit.HighPerformance/Properties/Polyfills/DoesNotReturnAttribute.cs
ovska: Good point, I've removed the attribute.

ovska:
Thanks for the review, I've fixed the nits!

tests/CommunityToolkit.HighPerformance.UnitTests/Extensions/Test_ArrayPoolExtensions.cs
ovska: I've added a test for EnsureCapacity after Rent with the same size; not sure if it's worth it to test more precisely as the exact size of the rented array isn't carved in stone. Let me know if you think of something.

CommunityToolkit.HighPerformance/Extensions/ArrayPoolExtensions.cs
```diff
@@ -53,4 +53,42 @@ public static void Resize<T>(this ArrayPool<T> pool, [NotNull] ref T[]? array, i
...
+        {
+            // Ensure rent succeeds before returning the original array to the pool
+            T[] newArray = pool.Rent(capacity);
+
+            pool.Return(array, clearArray);
+
+            array = newArray;
+        }
+
+        static void ThrowArgumentOutOfRangeExceptionForNegativeCapacity()
```
Sergio0694, line 89: Nit: due to Roslyn not optimizing this, using a local function here causes an unnecessary generic parameter being introduced here ([see here](https://github.com/dotnet/roslyn/issues/55800#issuecomment-1092965326)). Could you move this throw helper to a private static method in this class, outside of this method, and also add full XML docs on it while at it? You can see here as a reference:

https://github.com/CommunityToolkit/dotnet/blob/f3bd3d8b72615f1c1a5cd02876518ef01de518e5/CommunityToolkit.HighPerformance/Memory/Internals/ThrowHelper.cs#L74-L80

Sergio0694 (changes requested):
Looks great! Thank you for following the guide and force pushing as well 😄
Left one minor nit (not your fault, just Roslyn not being really helpful here).

Sergio0694 (approved):
LGTM, thanks! 🚀

PR #220, opened 2022-04-17
Dynamically Pack MVVM SourceGen project outputs

### LINKED ISSUE #219: [Build] Packing `SourceGen` projects fail when using a custom `OutputPath`s
## Problem

When building and packing the MVVM `SourceGenerator` projects, pack fails to find the output artifacts when we use custom output paths. This is observed in #96 when I tried to use a common output path for building the entire solution. I could update the current method to find the artifacts **but it feels hacky and not integrated with the build pipeline**.

Thus, any change like changing the output paths not just in the build definition but also changing via terminal invocation (_say `dotnet build -o path_to_out_dir`_) will definitely break the build and it won't pack those files. **Worst of all, sometimes, the build pipeline won't fail.** So, we won't know if the MVVM package was properly packed.

## To Reproduce

1. Clone the repo or clean it 
   `git clean -Xdf`
2. Build MVVM SourceGenerator project (_with a different output path_).
   `dotnet build CommunityToolkit.Mvvm.SourceGenerators -o bin_dir`
3. Pack the MVVM project (_with a different output path_).
   `dotnet pack CommunityToolkit.Mvvm -o pkg_dir`

## Result

| Expected | Actual |
|      ---      |    ---    |
| Build and pack should have succeeded and the outputs of the MVVM SourceGenerator project should be in the target path of NuGet package generated from the MVVM parent project. | Pack in the final step fails (_in a clean repo but succeeds when older artifacts are present_). When it does succeed, we either find the older artifacts or no artifacts in the NuGet package of the parent MVVM project. |

## Solution

Get the build outputs dynamically via `GetBuildOutputs` MSBuild target mirroring NuGet's Pack targets and add it to pack via `_PackageFiles ` internal MSBuild item used by NuGet's Pack targets. This is brittle only if @NuGet team decides to modify/remove `_PackageFiles` which is highly unlikely.

One interesting side-effect of this implementation is that when we decide to multi-target, we can easily modify our solution to be more resilient without having to tackle the problem again since we already output the artifacts by TFM. ___This should save hours of headbutting against the wall___ 😉!!

Nirmal4G:
**Closes #219**

Previously, static output path to the MVVM SourceGen assembly was used to pack the MVVM project.
This leads to error when OutputPath was updated dynamically when testing a new feature (see #96).

So, here, we'll update the build so that the SourceGen build outputs will be dynamically packed. We first get the build outputs dynamically via `GetBuildOutputs` MSBuild target mirroring NuGet's Pack targets. Then, we add it to the package via `_PackageFiles ` (_an internal MSBuild item used by NuGet's Pack targets_). This is brittle only if @NuGet team decides to modify/remove `_PackageFiles` which is highly unlikely.

## PR Checklist

- [x] Created a `feature` branch in your fork
- [x] PR doesn't include merge commits (**always rebase on top of our main**)

## Notes

_Also, this is brittle only if @NuGet team decides to modify/remove `_PackageFiles` which is highly unlikely._

One interesting side-effect of this implementation is that when we decide to multi-target, we can easily modify our solution to be more resilient without having to tackle the problem again since we already output the artifacts by TFM. ___This should save hours of headbutting against the wall___ 😉!!


Nirmal4G:
Please merge this only after merging #85 as it depends on some changes from that PR. All the 3 PRs are correlated since the changes cascade into the other.

PR #251, opened 2022-05-11
Ensure `Task` is `await`'d in `AsyncRelayCommand` and `AsyncRelayCommand<T>`

brminnick:
### Description

This PR ensures that `ICommand.Execute(object parameter)` and `IAsyncRelayCommand<T>.Execute(T parameter)` both `await` the execution of `Func<Task>? execute`.

It removes the bad practice of discarding a `Task`:
```cs
// Never discard a `Task`
_ = Task.Delay(100);

// Always await every `Task`
await Task.Delay(100);
```

I've added unit tests demonstrating why adding the `await` keyword is critical; the tests will fail if the updates in this PR to `AsyncRelayCommand` and `AsyncRelayCommand<T>` are removed.

### Explanation

Discarding a `Task` is a bad practice and should _never_ be done. 

Long story short, the compiler turns every `Task` into an `IAsyncStateMachine` in which a try/catch block is used to catch any `Exception` thrown inside of the `Task`.

If an `Exception` _is_ thrown inside of a `Task`, it is always re-thrown when the `await` keyword is used. If the `await` keyword is not used, the `Exception` is never re-thrown and is instead disposed when the `Task` is disposed, leading to bugs + race conditions.

I explain this in much more detail here in my AsyncAwaitBestPractices repo: https://github.com/brminnick/asyncawaitbestpractices#explaination

tl;dr A non-awaited Task doesn't rethrow Exceptions

## PR Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -232,3 +232,6 @@ msbuild.binlog
 # TAEF Log output
 WexLogFileOutput
 *.wtl
+
+#macOS Temporary Files
+*.DS_Store
--- a/CommunityToolkit.Mvvm/Input/AsyncRelayCommand.cs
+++ b/CommunityToolkit.Mvvm/Input/AsyncRelayCommand.cs
@@ -277,9 +277,9 @@ public sealed class AsyncRelayCommand : IAsyncRelayCommand, ICancellationAwareCo
     }
 
     /// <inheritdoc/>
-    public void Execute(object? parameter)
+    public async void Execute(object? parameter)
     {
-        _ = ExecuteAsync(parameter);
+        await ExecuteAsync(parameter).ConfigureAwait(false);
     }
 
     /// <inheritdoc/>
--- a/CommunityToolkit.Mvvm/Input/AsyncRelayCommand{T}.cs
+++ b/CommunityToolkit.Mvvm/Input/AsyncRelayCommand{T}.cs
@@ -214,7 +214,7 @@ public sealed class AsyncRelayCommand<T> : IAsyncRelayCommand<T>, ICancellationA
                 {
                     @this.PropertyChanged?.Invoke(@this, AsyncRelayCommand.ExecutionTaskChangedEventArgs);
                     @this.PropertyChanged?.Invoke(@this, AsyncRelayCommand.IsRunningChangedEventArgs);
-                    
+
                     if (@this.cancellationTokenSource is not null)
                     {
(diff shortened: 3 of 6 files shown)


brminnick:
@Sergio0694 Could you help me understand the Unit Test that is failing, `ThrowingTaskBubblesToUnobservedTaskException`?

My hunch/recommendation is that we should remove this test now that this PR ensures that we won't have an unobserved task exception. However, I wanted to confirm this with you first before removing it.

Sergio0694:
Hey @brminnick! 👋

I'm afraid the change in this PR is something that's just specifically against a behavior that was by design, so we can't merge this. The `Execute` methods were ignoring the tasks on purpose, as `ICommand` doesn't really have a concept of asynchronous operations, and we specifically didn't want to use `async void` and have the whole caller thread crash (ie. the whole application, as you'd usually invoke that method from the UI thread) if an operation failed. The way async commands are setup, the operation is only started when calling `Execute`, and then it can be monitored by the `ExecutionTask` property, so that you can eg. display some UI bit if an operation fails or something. The test that is failing is one that is specifically testing that calling `Execute` when the wrapped method does throw will _not_ crash the caller context, and also that if nobody accesses the `Task.Exception` property from `ExecutionTask`, that exception will bubble up to [TaskScheduler.UnobservedTaskException](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskscheduler.unobservedtaskexception), which is crucial to enable centralized logging of exceptions (as otherwise those exceptions could just be lost). The change in this PR is both breaking the guarantee that `Execute` will not crash the caller thread if the asynchronous operation fails (which then partially breaks the whole point of commands exposing the task), as well as the whole exception propagation system I mentioned 🥲

> "It removes the bad practice of discarding a `Task`"

I think this is where there might be a misunderstanding. That might often be bad practice, but here discarding the task is done intentionally and for a very specific reason, and the rest of the async command implementation is relying on this. The task is not _actually_ being ignored there (which would be bad, sure), but it's being assigned to `ExecutionTask` to allow for monitoring and for accessing it later on. It's not _really_ being discarded at all, it's just not being used at that specific call site 😊

michael-hawker:
🦙 Related to #22

brminnick:
Work for this bug is being solved in https://github.com/CommunityToolkit/dotnet/issues/262

PR #266, opened 2022-05-24
Asynchronous implementation of `MemorySource<TSource>.CopyToAsync`

### LINKED ISSUE #265: Asynchronous implementation of `MemorySource<TSource>.CopyToAsync`
### Overview

Currently, the stream returned by `AsStream()` implements `CopyToAsync` which calls the provided `CopyTo` implementation, which calls `destination.Write`. This causes unexpected synchronous I/O.

Due to the synchronous I/O, an exception is thrown if this method is used in ASP.NET Core to write to the response stream:

```
System.InvalidOperationException: Synchronous operations are disallowed. Call WriteAsync or set AllowSynchronousIO to true instead.
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpResponseStream.Write(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.Stream.Write(ReadOnlySpan`1 buffer)
   at Microsoft.Toolkit.HighPerformance.Streams.MemoryStream`1.CopyTo(Stream destination, Int32 bufferSize)
   at Microsoft.Toolkit.HighPerformance.Streams.MemoryStream`1.CopyToAsync(Stream destination, Int32 bufferSize, CancellationToken cancellationToken)
```

### API breakdown

**Note: These are internal APIs**

In order to implement this in a performant way, I'm thinking of adding a property `Memory` to `ISpanOwner`:
```csharp
internal interface ISpanOwner
{
    /// <summary>
    /// Gets a <see cref="Memory{T}"/> instance wrapping the underlying memory area.
    /// </summary>
    Memory<byte> Memory { get; }
}
```

And then implementing `CopyToAsync` like this:
```csharp
partial class MemoryStream<TSource>
{
    /// <inheritdoc/>
    public sealed override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
    {
        if (cancellationToken.IsCancellationRequested)
        {
            return Task.FromCanceled(cancellationToken);
        }

        try
        {
            MemoryStream.ValidateDisposed(this.disposed);

            Memory<byte> source = this.source.Memory.Slice(this.position);

            this.position += source.Length;

            return destination.WriteAsync(source, cancellationToken).AsTask();
        }
        catch (OperationCanceledException e)
        {
            return Task.FromCanceled(e.CancellationToken);
        }
        catch (Exception e)
        {
            return Task.FromException(e);
        }
    }
}
```

### Usage example

```csharp
public async Task ProcessRequestAsync(HttpContext context)
{
    var memory = await _service.GetDataAsync();
    var stream = memory.AsStream();
    await stream.CopyToAsync(context.Response.Body);
}
```

### Breaking change?

No

### Alternatives

Not applicable

### Additional context

Wasn't sure at first if this counted as a bug, upon closer inspection it looks a bit more likely, so I may have chosen the wrong template.

### Help us help you

OronDF343:
**Closes #265**

This PR replaces the implementation of `MemorySource<TSource>.CopyToAsync` with one that calls `destination.WriteAsync` instead of `destination.Write`. See linked issue for motivation.
To support this change, `ISpanOwner.Memory` has been added and implemented. Feedback appreciated.

## PR Checklist

## Other information

Please advise if tests are applicable for these changes.

### DIFF
--- a/CommunityToolkit.HighPerformance/Streams/Sources/ArrayOwner.cs
+++ b/CommunityToolkit.HighPerformance/Streams/Sources/ArrayOwner.cs
@@ -79,9 +79,6 @@ internal readonly struct ArrayOwner : ISpanOwner
     public Memory<byte> Memory
     {
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        get
-        {
-            return this.array.AsMemory(this.offset, this.length);
-        }
+        get => this.array.AsMemory(this.offset, this.length);
     }
 }
--- a/CommunityToolkit.HighPerformance/Streams/Sources/MemoryManagerOwner.cs
+++ b/CommunityToolkit.HighPerformance/Streams/Sources/MemoryManagerOwner.cs
@@ -66,9 +66,6 @@ internal readonly struct MemoryManagerOwner : ISpanOwner
     public Memory<byte> Memory
     {
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        get
-        {
-            return this.memoryManager.Memory.Slice(this.offset, this.length);
-        }
+        get => this.memoryManager.Memory.Slice(this.offset, this.length);
     }
 }


CommunityToolkit.HighPerformance/Streams/MemoryStream{TSource}.cs
```diff
@@ -119,9 +119,13 @@ public sealed override Task CopyToAsync(Stream destination, int bufferSize, Canc
...
         {
-            CopyTo(destination, bufferSize);
+            MemoryStream.ValidateDisposed(this.disposed);
 
-            return Task.CompletedTask;
+            Memory<byte> source = this.source.Memory.Slice(this.position);
+
+            this.position += source.Length;
+
+            return destination.WriteAsync(source, cancellationToken).AsTask();
```
gfoidl, line 128: Should the order be swapped?
If any exception occurs, then position won't be updated, which is what's expected?

OronDF343: I implemented this based on the existing `CopyTo` implementation, which itself seems to be inconsistent with the `Read` and `Write` overloads. Should I fix the original within this PR as well?

Sergio0694: The order there is following what `MemoryStream` does in the BCL ([see here](https://github.com/dotnet/runtime/blob/7adbc10dfb865fd5ecc37e309caa2d955b6061b0/src/libraries/System.Private.CoreLib/src/System/IO/MemoryStream.cs#L368)). The `Read` and `Write` APIs read/write first and then update the position, whereas `CopyTo` and `CopyToAsync` first update the position and then defer to the destination stream. We're just following suit with that same logic.

CommunityToolkit.HighPerformance/Streams/Sources/ArrayOwner.cs
```diff
@@ -74,4 +74,14 @@ public Span<byte> Span
...
     }
+
+    /// <inheritdoc/>
+    public Memory<byte> Memory
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get
+        {
+            return this.array.AsMemory(this.offset, this.length);
+        }
```
Sergio0694, line 85: Nit: use a one liner.
```suggestion
        get => this.array.AsMemory(this.offset, this.length);
```

CommunityToolkit.HighPerformance/Streams/Sources/MemoryManagerOwner.cs
```diff
@@ -61,4 +61,14 @@ public Span<byte> Span
...
     }
+
+    /// <inheritdoc/>
+    public Memory<byte> Memory
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get
+        {
+            return this.memoryManager.Memory.Slice(this.offset, this.length);
+        }
```
Sergio0694, line 72: Use a one liner here as well.
```suggestion
        get => this.memoryManager.Memory.Slice(this.offset, this.length);
```

Sergio0694 (changes requested):
Could you add some unit tests too? 🙂

OronDF343:
I've added basic unit tests

PR #279, opened 2022-05-27
Add support for "On" prefix for command methods

### LINKED ISSUE #162: [Enhancement Request]: allow method prefixes for `ICommand`
I would previously name methods i was using for a command with either `On<command>` or `On<command>Async`.  Like `ObservableProperty` that allows `_` or `m_` prefixes, I think it would be nice to allow method prefixes.

Sergio0694:
**Closes #162**

## PR Checklist

- [ ] New component

### DIFF
--- a/CommunityToolkit.Mvvm.SourceGenerators/Input/ICommandGenerator.Execute.cs
+++ b/CommunityToolkit.Mvvm.SourceGenerators/Input/ICommandGenerator.Execute.cs
@@ -350,6 +350,17 @@ partial class ICommandGenerator
         {
             string propertyName = methodSymbol.Name;
 
+            // Strip the "On" prefix, if present. Only do this if the name is longer than 2 characters,
+            // and if the 3rd character is not a lowercase letter. This is needed to avoid accidentally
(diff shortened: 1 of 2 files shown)


pwshdevops:
When upgrading from prior to Preview 4, if you have commands which already start with an "On" prefix, your commands will be updated without the prefix. ViewModel command binding references ought to be updated accordingly.

Is this an artifact of this or #288 ?

Sergio0694:
@pwshdevops Not sure I understand what you mean. The changes in #288 are completely unrelated to this.
If you upgrade from an earlier preview, then your code will just break, and that's just part of using a preview 😄

pwshdevops:
> @pwshdevops Not sure I understand what you mean. The changes in #288 are completely unrelated to this. If you upgrade from an earlier preview, then your code will just break, and that's just part of using a preview 😄

You're right, I apologize. I had assumed there was more source generation magic behind #288 than simply renaming the attributes.

My comment was a awful attempt to communicate that the update will rename the generated command if you have an "On" prefix -- should anyone be scratching their heads, like I was.

PR #281, opened 2022-05-28
Remove temporary strings in source generators formatting

Sergio0694:
See https://github.com/dotnet/roslyn/issues/61506 and https://github.com/Sergio0694/ComputeSharp/pull/297.

cc. @Neme12

## PR Checklist

- [ ] New component

### DIFF
--- a/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableRecipientGenerator.cs
+++ b/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableRecipientGenerator.cs
@@ -117,19 +117,18 @@ public sealed class ObservableRecipientGenerator : TransitiveMembersGenerator<Ob
         // If the target type has no constructors, generate constructors as well
         if (!info.HasExplicitConstructors)
         {
-            foreach (ConstructorDeclarationSyntax ctor in memberDeclarations.OfType<ConstructorDeclarationSyntax>())
+            foreach (ConstructorDeclarationSyntax originalConstructor in memberDeclarations.OfType<ConstructorDeclarationSyntax>())
             {
-                string text = ctor.NormalizeWhitespace().ToFullString();
-                string replaced = text.Replace("ObservableRecipient", info.TypeName);
+                ConstructorDeclarationSyntax modifiedConstructor = originalConstructor.WithIdentifier(Identifier(info.TypeName));
 
                 // Adjust the visibility of the constructors based on whether the target type is abstract.
                 // If that is not the case, the constructors have to be declared as public and not protected.
                 if (!info.IsAbstract)
                 {
-                    replaced = replaced.Replace("protected", "public");
+                    modifiedConstructor = modifiedConstructor.WithModifiers(TokenList(Token(SyntaxKind.PublicKeyword)));
                 }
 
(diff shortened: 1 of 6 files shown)


CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableRecipientGenerator.cs
```diff
@@ -117,19 +117,18 @@ protected override ImmutableArray<MemberDeclarationSyntax> FilterDeclaredMembers
...
-                string text = ctor.NormalizeWhitespace().ToFullString();
-                string replaced = text.Replace("ObservableRecipient", info.TypeName);
+                ConstructorDeclarationSyntax modifiedConstructor = originalConstructor.WithIdentifier(Identifier(info.TypeName));
 
                 // Adjust the visibility of the constructors based on whether the target type is abstract.
                 // If that is not the case, the constructors have to be declared as public and not protected.
                 if (!info.IsAbstract)
                 {
-                    replaced = replaced.Replace("protected", "public");
+                    modifiedConstructor = modifiedConstructor.WithModifiers(TokenList(Token(SyntaxKind.PublicKeyword)));
```
Neme12, line 128: What if the constructor has other modifiers as well, like `unsafe`? This would remove all of them. Or is this constructor not from user code but generated by the generator and it's guaranteed that it won't have other modifiers?

CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableRecipientGenerator.cs
```diff
@@ -117,19 +117,18 @@ protected override ImmutableArray<MemberDeclarationSyntax> FilterDeclaredMembers
...
                 }
 
-                builder.Add((ConstructorDeclarationSyntax)ParseMemberDeclaration(replaced)!);
```
Neme12, line 132: Ugh it was being reparsed just to replace a token... I'm glad this is fixed 😄

CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableRecipientGenerator.cs
Neme12: If preserving other modifiers is needed, try something like this:
```c#
modifiedConstructor = modifiedConstructor.WithModifiers(
    modifiedConstructor.Modifiers.Replace(
        modifiedConstructor.Modifiers.First(token => token.Kind() is SyntaxKind.ProtectedKeyword),
        SyntaxFactory.Token(SyntaxKind.PublicKeyword)));
```

Sergio0694: The syntax tree is from the `ObservableRecipient.cs` embedded resource file, so yeah we can assume that `protected` is the only modifier that could possibly be there. These trees are not provided by code parsed from consumers 🙂

CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableRecipientGenerator.cs
Sergio0694: Yup, looking back at this it was indeed quite bad ahah

PR #289, opened 2022-06-01
Strong-name sign assemblies

### LINKED ISSUE #254: Latest preview assemblies are not signed
### Describe the bug

Since v8.0.0-preview.1, the toolkit DLL is no longer strong name signed and cannot be loaded by .NETFramework libraries/executables that are strong named.

### Regression

Yes. v7.1.2 is signed.

### Steps to reproduce

```text
1. Have a WPF app that targets both `net48` and `net6.0-windows` and is strong named.
2. Try to use the latest preview.
3. Works for NET 6.0 (because strong names no longer matter there) but not for net48.
```

### Expected behavior

CommunityToolkit.Mvvm.dll should be signed with the toolkit.snk that's in the root directory.

### Screenshots

v7.1.2  

image

v8.0.0-preview

image

### IDE and version

VS 2022 Preview

### IDE version

17.3-p1

### Nuget packages

- [ ] CommunityToolkit.Common
- [X] CommunityToolkit.Mvvm (aka MVVM Toolkit)

### Nuget package version(s)

v8.0.0-preview.3

### Additional context

_No response_

### Help us help you

InKahootz:
**Closes #254**

Enable strong-name signing for all assemblies. This fixes a regression introduced in the repo migration. v7 was strong-named signed with the key that existed in the root of the repo.

## PR Checklist

## Other information

N/A


Sergio0694 (changes requested):
Can you also double check with a package from our PR feed to ensure it solves the issue for you? 🙂

InKahootz:
See my discussion inside the original issue. You cannot use `IsCoreProject` (which is exactly what I did at first 😅) because InternalsVisibleTo doesn't work without a strong name on the unit test project.

Sergio0694:
Ok but then I'm confused as to how can the projects in the WCT use `[IVT]` without signing the tests though 🤔
@azchohfi would you happen to know how that's setup there?

InKahootz:
I can confirm I am seeing signed assemblies from the PR artifacts

image

InKahootz:
> Uh... Why is the company ".NET Foundation" though. It should be "Microsoft". Anyway that's a separate point, but @michael-hawker we should clarify and align all those line items I think 🙂

That would be handled here: https://github.com/CommunityToolkit/dotnet/blob/157436d2592d3487041aaea066c6bc161e95de26/build/Community.Toolkit.Common.props#L10

I can open another issue for that one

azchohfi:
> @azchohfi would you happen to know how that's setup there?

I honestly don't remember... I don't remember that the tests had to be signed. If the tests were signed, then yes, everything they reference should be signed, but not the other way around, right?

InKahootz:
> but not the other way around, right?

Not always. `IVT` attribute requires that the friend assembly be signed as well if the one using the attribute is signed.

In one case, the `HighPerformance.UnitTests` tests must be signed because it's being given friend access via the HighPerformance project.

You can try to build this branch with this instead but you will encounter a few [CS1726](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1726?f1url=%3FappId%3Droslyn%26k%3Dk(CS1726)) build errors.
```xml
  <PropertyGroup Condition=" '$(IsCoreProject)' == 'true' ">
    <SignAssembly>true</SignAssembly>
    <AssemblyOriginatorKeyFile>$(RepositoryDirectory)toolkit.snk</AssemblyOriginatorKeyFile>
  </PropertyGroup>
```

CommunityToolkit.HighPerformance/Properties/AssemblyInfo.cs
```diff
@@ -12,5 +12,5 @@
 
 // We need to test the RuntimeHelpers polyfills on applicable runtimes
 #if !NETSTANDARD2_1_OR_GREATER
-[assembly: InternalsVisibleTo("CommunityToolkit.HighPerformance.UnitTests")]
+[assembly: InternalsVisibleTo("CommunityToolkit.HighPerformance.UnitTests, PublicKey=002400000480000094000000060200000024000052534131000400000100010041753af735ae6140c9508567666c51c6ab929806adb0d210694b30ab142a060237bc741f9682e7d8d4310364b4bba4ee89cc9d3d5ce7e5583587e8ea44dca09977996582875e71fb54fa7b170798d853d5d8010b07219633bdb761d01ac924da44576d6180cdceae537973982bb461c541541d58417a3794e34f45e6f2d129e2")]
```
Sergio0694, line 15: Nit: can you make the keys be uppercase?

Sergio0694 (approved):
Minor nit, other than that LGTM! :shipit:

Thanks!

CommunityToolkit.HighPerformance/Properties/AssemblyInfo.cs
```diff
@@ -12,5 +12,5 @@
 
 // We need to test the RuntimeHelpers polyfills on applicable runtimes
 #if !NETSTANDARD2_1_OR_GREATER
-[assembly: InternalsVisibleTo("CommunityToolkit.HighPerformance.UnitTests")]
+[assembly: InternalsVisibleTo("CommunityToolkit.HighPerformance.UnitTests, publickey=002400000480000094000000060200000024000052534131000400000100010041753AF735AE6140C9508567666C51C6AB929806ADB0D210694B30AB142A060237BC741F9682E7D8D4310364B4BBA4EE89CC9D3D5CE7E5583587E8EA44DCA09977996582875E71FB54FA7B170798D853D5D8010B07219633BDB761D01AC924DA44576D6180CDCEAE537973982BB461C541541D58417A3794E34F45E6F2D129E2")]
 #endif
```
Nirmal4G, line 16: Do we need to guard it so that it only exists in CI?

CommunityToolkit.Mvvm/Properties/AssemblyInfo.cs
```diff
@@ -7,4 +7,4 @@
 // See notes in CommunityToolkit.HighPerformance for more info
 [module: SkipLocalsInit]
 
-[assembly: InternalsVisibleTo("CommunityToolkit.Mvvm.Internals.UnitTests")]
\ No newline at end of file
+[assembly: InternalsVisibleTo("CommunityToolkit.Mvvm.Internals.UnitTests, PublicKey=002400000480000094000000060200000024000052534131000400000100010041753AF735AE6140C9508567666C51C6AB929806ADB0D210694B30AB142A060237BC741F9682E7D8D4310364B4BBA4EE89CC9D3D5CE7E5583587E8EA44DCA09977996582875E71FB54FA7B170798D853D5D8010B07219633BDB761D01AC924DA44576D6180CDCEAE537973982BB461C541541D58417A3794E34F45E6F2D129E2")]
```
Nirmal4G, line 10: Here too? These are shipping assemblies!! Do we need this in final binary?

InKahootz: It's very normal for these to exist in a shipping binary. Go check almost any major NuGet package.

Nirmal4G: Just because every package out there does it doesn't mean we have too! I was just wondering if it's best practice!

InKahootz: It's not like you can drop this attribute anyhow. The whole reason for the PR is to bring signing back up to parity with v7. Read the previous discussion on the PR about `IVT` errors.

Nirmal4G: I actually welcome this change but I didn't want to leave the attribute in the release. I could guard it but CI then would require two builds, one for tests and one for release.

There's no way to do it otherwise, so, I didn't do it in my refactor PR!

Nirmal4G (approved):

Nirmal4G:
**PSA** This is okay as it is but I found this -> dotnet/sdk#3439

**FYI** We can now add `InternalsVisibleTo` in the project file!

```xml
<!-- Common.props (or in Azure Pipelines) -->
<PublicKey>002400000480000094000000060200000024000052534131000400000100010041753AF735AE6140C9508567666C51C6AB929806ADB0D210694B30AB142A060237BC741F9682E7D8D4310364B4BBA4EE89CC9D3D5CE7E5583587E8EA44DCA09977996582875E71FB54FA7B170798D853D5D8010B07219633BDB761D01AC924DA44576D6180CDCEAE537973982BB461C541541D58417A3794E34F45E6F2D129E2</PublicKey>

<!-- CommunityToolkit.Mvvm -->
<InternalVisibleTo Include="CommunityToolkit.Mvvm.Internals.UnitTests" PublicKey="$(PublicKey)"/>

<!-- CommunityToolkit.HighPerformance -->
<InternalVisibleTo Include="CommunityToolkit.HighPerformance.UnitTests" PublicKey="$(PublicKey)"/>
```

InKahootz:
I'm aware of the project property. I'm keeping it simple and not modifying more than what is necessary to fix the opened issue.

Simple PRs are less demanding to review.

Sergio0694:
Thank you @InKahootz! 😄

PR #290, opened 2022-06-02
Update Company and Authors metadata items

Sergio0694:
This PR updates the company and authors metadata tags to be "Microsoft".

### DIFF
--- a/build/Community.Toolkit.Common.props
+++ b/build/Community.Toolkit.Common.props
@@ -1,8 +1,8 @@
 <Project>
 
   <PropertyGroup>
-    <Company>.NET Foundation</Company>
-    <Authors>Microsoft.Toolkit</Authors>
+    <Company>Microsoft</Company>
+    <Authors>Microsoft</Authors>
     <Product>.NET Community Toolkit</Product>
     <CommonTags>dotnet;Community;Toolkit</CommonTags>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>


build/Community.Toolkit.Common.props
```diff
@@ -1,8 +1,8 @@
 <Project>
 
   <PropertyGroup>
-    <Company>.NET Foundation</Company>
-    <Authors>Microsoft.Toolkit</Authors>
+    <Company>Microsoft</Company>
+    <Authors>Microsoft</Authors>
```
Sergio0694, line 5: @azchohfi just to confirm, would you happen to know where this `<Authors>` tag is used anyway?

It's not showing up in the assembly attributes, like `<Company>` is (which maps to `[AssemblyCompany]`. And I'm pretty sure it's unrelated from the actual publisher name that's used on NuGet (which would remain "Microsoft.Toolkit") 🤔

Nirmal4G: For NuGet, `Owners` has been depreciated. `Authors` map to both NuGet package authors and owners now. If we're under `.NET Foundation`, we should use that.

@michael-hawker You said we move to `CommunityToolkit` org for the NuGet instead of `Microsoft.Toolkit`, is it done?

michael-hawker: @Nirmal4G the rename there is for the package identity and not related to our org name here or the authorship of the package account on NuGet.

My original understanding was that `Authors` maps to the NuGet account holder names for the packages, and I thought what tied it to our account (though it's probably the key we use for uploading that does that instead). Though according to this [doc here](https://docs.microsoft.com/en-us/nuget/create-packages/package-authoring-best-practices#authors) it seems like we may just be able to use what we want?

> ✔️ DO use the author field for your or your organization's "pretty name."
>
> For example, if my NuGet.org username is "jdoe" then using "Jane Doe" for the author field may help consumers recognize me as an author. If my organization's NuGet.org username is "ContosoToolkit" then using "Contoso Corporation" may be more recognizable and inspire more consumer trust.

In which case we could try and put `Microsoft, .NET Foundation, Community Toolkit`? If you want to be a guinea pig, we can see what happens with the next 8.0 preview? Though I'd still get @azchohfi's thoughts first. 😋

Sergio0694: Good to hear it's separate and we can update it independently! 😄
I don't think we should put .NET Foundation or CommunityToolkit there, just Microsoft. It's the same with, say, .NET itself. It's still part of the .NET Foundation, but the author is still Microsoft. Same goes for all community contributions: contributors will sign the CLA and contribute code under MIT and the .NET Foundation licensing, but that's separate to what the author or company is for the package. Especially with so many people not understanding that this is a 1st party library (even more so since the org rename), I think we should go with Microsoft in both fields.

Sergio0694:
Merging this and then publishing 8.0.0 Preview 4, and we'll see if everything works out properly then 😄

Sergio0694:
Merging this and then publishing 8.0.0 Preview 4, and we'll see if everything works out properly then 😄

PR #292, opened 2022-06-04
Add FlowExceptionsToTaskScheduler command option

### LINKED ISSUE #262: Ability to customize exception flow in AsyncRelayCommand types
### Overview

This issue is about discussing ways to improve how `AsyncRelayCommand` types handle exceptions. Specifically, as discussed with @brminnick and as people have brought up in the past as well (eg. #22, #175, #251, etc.), the way async relay commands currently work is that if you invoke them through `Execute` (ie. through a synchronous API), the `Task` object is not awaited. This is **by design**, as it allows exceptions to be observable through the `ExecutionTask` property afterwards, or to otherwise bubble up to `TaskScheduler.UnobservedTaskException` to enable centralized logging. Currently, if someone wants to await a task, `ExecuteAsync` needs to be called instead. This also has the characteristic that calling `Execute` will not crash the app if the command fails.

Now, there are cases where all this is desireable, but there might also be cases where developers would want the exception to just be reported immediately and crash the app, for instance to more easily spot errors and to avoid the app getting into an inconsistent state, in case an exception was not planned to be raised from the wrapped methods at all.

Because of this, we're thinking about ways we could expose this additional flexibility for consumers.

This issue is meant to be a place to discuss this and the various API proposals related to this.

cc. @michael-hawker @Arlodotexe 

### API breakdown

One possible way to go could be to introduce something like this:

```csharp
namespace CommunityToolkit.Mvvm.Input;

[Flags]
public enum AsyncRelayCommandOptions
{
    Default,
    AllowConcurrentExecution,
    FlowExceptionsToThreadPool,
    FlowExceptionsToSynchronizationContext,
    // Potentially more options in the future
}
```

This would then be used both in constructors for `AsyncRelayCommand`, as well as `[ICommand]`.

Worth noting, this would _replace_ the current `bool allowConcurrentExecutions` parameters in the constructors, as instead we would just take an `AsyncRelayCommandOptions options` parameter to specify all options. This way, consumers would have a single way to customize the behavior of the async command on all supported aspects.

As far as the source generator approach goes, we could do something like:

```csharp
// Before
[ICommand(AllowConcurrentExecutions = true)]
private async Task DoSomethingAsync()
{
    // ...
}

// After
[ICommand(Options = AsyncRelayCommandOptions.AllowConcurrentExecutions)]
private async Task DoSomethingAsync()
{
    // ...
}
```

This is a bit more verbose than before. It might be ok, but I'm also open to suggestions.

Overall, using something like this would give us the following benefits:

- Everyone would be happy, as we could support all scenarios we wanted
- Having some options would give us room to potentially add more flags in the future without breaking or needing new APIs

### Open questions

Should the default be the same behavior as today, or should the default be to rethrow in some specific way?

### Alternatives

There isn't really a good alternative to have the app crash today. I mean you could do that but it'd be incredibly verbose, as you'd need to register a handler for `PropertyChanged` on the command, check whether the task changed, get it, check if it faulted, then post to the synchronization context you need and from there re-throw the exceptions. Not really ideal.

Sergio0694:
**Closes #262**
**Picks up work in #251**

## API breakdown

```csharp
namespace CommunityToolkit.Mvvm.Input;

[Flags]
public enum AsyncRelayCommandOptions
{
    None,
    AllowConcurrentExecutions,
    FlowExceptionsToTaskScheduler
}

public sealed class RelayCommandAttribute
{
    public bool FlowExceptionsToTaskScheduler { get; init; }
}

// Updated all AsyncRelayCommand and AsyncRelayCommand<T> constructors too
```

> ⚠️ **Breaking change:** the default behavior for async commands is now to await and throw exceptions.

## PR Checklist

- [ ] New component


brminnick (approved):
Awesome! Thanks Sergio!!

My only comment (which isn't pertinent to this PR) is to combine some of the overlapping code between `AsyncRelayCommand` and `AsyncRelayCommand<T>` into a base class, something like `BaseAsyncRelayCommand`.

Here's an example of how I handled it when I created an AsyncCommand back in the day:
- [`BaseAsyncCommand`](https://github.com/brminnick/AsyncAwaitBestPractices/blob/main/Src/AsyncAwaitBestPractices.MVVM/AsyncCommand/BaseAsyncCommand.shared.cs)
- [`AsyncCommand`](https://github.com/brminnick/AsyncAwaitBestPractices/blob/b521300f01cce44eb4bff681f024e8a6b6cd59da/Src/AsyncAwaitBestPractices.MVVM/AsyncCommand/AsyncCommand.shared.cs#L65)
- [`AsyncCommand<T>`](https://github.com/brminnick/AsyncAwaitBestPractices/blob/b521300f01cce44eb4bff681f024e8a6b6cd59da/Src/AsyncAwaitBestPractices.MVVM/AsyncCommand/AsyncCommand.shared.cs#L37)

CommunityToolkit.Mvvm/Input/AsyncRelayCommand.cs
```diff
@@ -323,4 +330,13 @@ public void Cancel()
...
         }
     }
+
+    /// <summary>
+    /// Awaits an input <see cref="Task"/> and throws an exception on the calling context, if the task fails.
+    /// </summary>
+    /// <param name="executionTask">The input <see cref="Task"/> instance to await.</param>
+    internal static async void AwaitAndThrowIfFailed(Task executionTask)
+    {
+        await executionTask;
```
pictos, line 340: can we add the `ConfigureAwait(false)` here?
```suggestion
        await executionTask.ConfigureAwait(false);
```

pictos (commented):
This looks great, really love the new `Options` I just have one suggestion

CommunityToolkit.Mvvm/Input/AsyncRelayCommand.cs
Sergio0694: I didn't add this on purpose to be honest. The idea is that if one of the main points of having exceptions be rethrown is to also help debugging things (other than not get the app into a weird invalid state of course), I feel like it'd be clearer for users to see the exceptions be thrown on the calling context, which would generally be the UI thread. This is also handled in special ways on many frameworks, versus exceptions being thrown on random thread pool threads. I know there's a slight overhead in this, but I feel like that's fine here given it's not like this would be used in a hot path like in some asynchronous code dealing with loops of web requests or anything like that 🤔

pictos: It's fine then. I thought that could be intentional, but not sure if with the `ConfigureAwait(false)` the exception will be thrown in another context. Thanks for the explanation (:

CommunityToolkit.Mvvm/Input/AsyncRelayCommand.cs
```diff
@@ -323,4 +330,13 @@ public void Cancel()
             PropertyChanged?.Invoke(this, IsCancellationRequestedChangedEventArgs);
         }
     }
+
+    /// <summary>
+    /// Awaits an input <see cref="Task"/> and throws an exception on the calling context, if the task fails.
+    /// </summary>
+    /// <param name="executionTask">The input <see cref="Task"/> instance to await.</param>
+    internal static async void AwaitAndThrowIfFailed(Task executionTask)
```
Arlodotexe, line 338: Can we add some notes regarding the need for `async void` here?

Arlodotexe (approved):

CommunityToolkit.Mvvm/Input/AsyncRelayCommand.cs
Sergio0694: Done in https://github.com/CommunityToolkit/dotnet/pull/292/commits/612055be6f69a057d6774f60905183828634a669.

PR #301, opened 2022-06-08
Improve exceptions for invalid command arguments

### LINKED ISSUE #296: Relay commands should throw ArgumentException and not InvalidCastException
### Describe the bug

When a generic relay command receives an input argument of an invalid type, a cast is attempted, resulting in an `InvalidCastException`. This is particularly hard to debug (see #282), and in general an exception type that should just never be thrown by user code (at least on purpose). We should validate the arguments and throw an `ArgumentException` instead, with a useful error message (eg. saying what type was used, what type was expected, what method is doing what, etc.).

### Regression

No, always been like this

### Steps to reproduce

1. Create a `RelayCommand<int>`
2. Call `command.Execute("Hello")

An `InvalidCastException` is thrown.
### Expected behavior

An `ArgumentException` is thrown with a useful error message.

Sergio0694:
**Closes #296**

This exception changes the generic commands to always thrown `ArgumentException` if the input is invalid.
This also includes a helpful exception message.
They no longer throw `NullReferenceException` or `InvalidCastException`.

## PR Checklist

### DIFF
--- a/CommunityToolkit.Mvvm/Input/RelayCommand{T}.cs
+++ b/CommunityToolkit.Mvvm/Input/RelayCommand{T}.cs
@@ -6,6 +6,7 @@
 // more info in ThirdPartyNotices.txt in the root of the project.
 
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 
 namespace CommunityToolkit.Mvvm.Input;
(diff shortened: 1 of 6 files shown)


CommunityToolkit.Mvvm/Input/RelayCommand{T}.cs
```diff
@@ -100,6 +107,66 @@ public void Execute(T? parameter)
...
+
+    /// <summary>
+    /// Throws an <see cref="ArgumentException"/> if an invalid command argument is used.
+    /// </summary>
+    /// <param name="parameter">The input parameter.</param>
+    /// <exception cref="ArgumentException">Thrown with an error message to give info on the invalid parameter.</exception>
+    [DoesNotReturn]
+    internal static void ThrowArgumentExceptionForInvalidCommandArgument(object? parameter)
+    {
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
Arlodotexe, line 159: Is `NoInlining` intended here?

Sergio0694: Yeah it's to make sure the JIT doesn't mess up handling the caller as a throw helper.

PR #309, opened 2022-06-11
Add "this." prefix to all field accesses

Sergio0694:
This PR adds the missing `this.` prefix to all leftover field accesses.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -62,8 +62,8 @@ dotnet_style_parentheses_in_other_binary_operators = always_for_clarity:silent
 dotnet_style_parentheses_in_other_operators = never_if_unnecessary:silent
 dotnet_style_parentheses_in_relational_binary_operators = always_for_clarity:silent
 
-# Modifier preferences
-dotnet_style_require_accessibility_modifiers = for_non_interface_members:warning
+# Modifier preferences (set to silent until https://github.com/dotnet/roslyn/issues/52904 is resolved)
+dotnet_style_require_accessibility_modifiers = for_non_interface_members:silent
 
 # Code block preferences
 dotnet_style_allow_multiple_blank_lines_experimental = false:warning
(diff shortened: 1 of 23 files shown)


Nirmal4G:
Seeing the code, we'll know whether it's a parameter from the method or a field and a property within the class/struct. Do we have to add these? It is visually distracting. IMO, it's best to remove all `this` qualifiers unless the names of the method and parameter are the same which is rare and compiler would certainly complain about it.

So, I think it is safe to remove all `this`! On that note, we should also come up with a definite way to name fields. Whether prefixing `_` underscore is required or not. There are a lot of inconsistencies here. I vote **NO** but some .NET projects use them!

Sergio0694:
Fields should just be named `lowerCamelCase`. The only reason why some use `_lowerCamelCase` is that they're from old classes we haven't refactored yet. I guess I might as well update those too while at it, since it's not a breaking change anyway.

> "Seeing the code, we'll know whether it's a parameter from the method or a field and a property within the class/struct. Do we have to add these? It is visually distracting."

Adding `this.` makes the code less error prone, as it avoids the risk of accidentally referencing a local or a parameter with the same name. Yes, you can manually add `this.` when there's a conflict, but that is precisely the point - making this mandatory means you don't have to rememeber, and you can't make mistakes. If there's `this`, it's 100% a field.

Nirmal4G:
> Yes, you can manually add `this`. when there's a conflict, but that is precisely the point - making `this` mandatory means you don't have to remember, and you can't make mistakes. If there's `this`, it's 100% a field.

Guess it makes sense! Still, I personally hate it. I just use different names for parameters, locals and fields.

Balkoth:
Theres [this](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md) directly from the runtime team.

Sergio0694:
That is not the official guidance for all C# developers, just the style convention the runtime uses 🙂

PR #310, opened 2022-06-11
Add missing trimming annotations

Sergio0694:
This PR adds two missing trimming annotations to a couple of APIs.

### DIFF
--- a/CommunityToolkit.Common/Extensions/TaskExtensions.cs
+++ b/CommunityToolkit.Common/Extensions/TaskExtensions.cs
@@ -2,6 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#if NET6_0_OR_GREATER
+using System.Diagnostics.CodeAnalysis;
+#endif
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


CommunityToolkit.Common/Extensions/TaskExtensions.cs
```diff
@@ -2,6 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#if NET6_0_OR_GREATER
+using System.Diagnostics.CodeAnalysis;
+#endif
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
```
Nirmal4G, line 10: Can we put all the conditional `using` below the top-level ones? If possible, with one line spacing!?

CommunityToolkit.Mvvm/ComponentModel/ObservableValidator.cs
```diff
@@ -802,6 +802,7 @@ private void ClearErrorsForProperty(string propertyName)
     /// </summary>
     /// <param name="propertyName">The target property name being validated.</param>
     /// <returns>The display name for the property.</returns>
+    [RequiresUnreferencedCode("The type of the current instance cannot be statically discovered.")]
```
Nirmal4G, line 805:
```suggestion
    [RequiresUnreferencedCode("The current instance type cannot be statically discovered.")]
```

How about this?

Sergio0694: This is a standard error message also used in many APIs in the BCL, I want it to be consistent.

CommunityToolkit.Common/Extensions/TaskExtensions.cs
Sergio0694: We're doing this in all other files, so just keeping it simple for consistency here.

Nirmal4G: Is the preferred style? If so, why? Almost all of the projects from .NET keep conditional `using` below the top-level ones.

Sergio0694: I'm not saying this is necessarily the preferred style, just the one we've used so far. This PR is just mean to add a missing annotation, so I wouldn't want to do additional unrelated refactoring here. We might want to change this in the future, but this PR is just not the right place to do that anyway, so that's why I'm leaving this bit as is for now 🙂

PR #321, opened 2022-06-21
Add unique prefixes to all .targets properties and items

Sergio0694:
This PR adds unique prefixes to all properties and items in the .targets file bundled with the MVVM Toolkit.
This ensures there can be no conflicts with other targets defining the same elements from other NuGet packages.

Essentially, this PR fixes this: https://github.com/Sergio0694/ComputeSharp/actions/runs/2523630796.

## PR Checklist

### DIFF
--- a/CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.targets
+++ b/CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.targets
@@ -1,17 +1,17 @@
 <Project>
 
   <!-- Get the analyzer from the CommunityToolkit.Mvvm NuGet package -->
-  <Target Name="_MVVMToolkitGatherAnalyzers">
+  <Target Name="MVVMToolkitGatherAnalyzers">
     <ItemGroup>
-      <_MVVMToolkitAnalyzer Include="@(Analyzer)" Condition="'%(Analyzer.NuGetPackageId)' == 'CommunityToolkit.Mvvm'" />
+      <MVVMToolkitAnalyzer Include="@(Analyzer)" Condition="'%(Analyzer.NuGetPackageId)' == 'CommunityToolkit.Mvvm'" />
     </ItemGroup>
   </Target>
 
   <!-- Remove the analyzer if using Roslyn 3.x (incremental generators require Roslyn 4.x) -->
-  <Target Name="_MVVMToolkitRemoveAnalyzersForRoslyn3"
+  <Target Name="MVVMToolkitRemoveAnalyzersForRoslyn3"
           Condition="'$(CSharpCoreTargetsPath)' != ''"
           AfterTargets="ResolvePackageDependenciesForBuild;ResolveNuGetPackageAssets"
-          DependsOnTargets="_MVVMToolkitGatherAnalyzers">
+          DependsOnTargets="MVVMToolkitGatherAnalyzers">
 
     <!-- Use the CSharpCoreTargetsPath property to find the version of the compiler we are using. This is the same mechanism
          MSBuild uses to find the compiler. We could check the assembly version for any compiler assembly (since they all have
@@ -32,7 +32,7 @@
 
     <!-- If the Roslyn version is < 4.0, disable the source generators -->
     <ItemGroup Condition ="'$(MVVMToolkitCurrentCompilerVersionIsNotNewEnough)' == 'true'">
-      <Analyzer Remove="@(_MVVMToolkitAnalyzer)"/>
+      <Analyzer Remove="@(MVVMToolkitAnalyzer)"/>
     </ItemGroup>
 
     <!-- If the source generators are disabled, also emit a warning. This would've been produced by MSBuild itself as well, but
@@ -42,14 +42,14 @@
   </Target>
 
   <!-- Remove the analyzer if Roslyn is missing -->
-  <Target Name="_MVVMToolkitRemoveAnalyzersForRosynNotFound"
+  <Target Name="MVVMToolkitRemoveAnalyzersForRosynNotFound"
           Condition="'$(CSharpCoreTargetsPath)' == ''"
           AfterTargets="ResolvePackageDependenciesForBuild;ResolveNuGetPackageAssets"
-          DependsOnTargets="_MVVMToolkitGatherAnalyzers">
+          DependsOnTargets="MVVMToolkitGatherAnalyzers">
 
     <!-- If no Roslyn assembly could be found, just remove the analyzer without emitting a warning -->
     <ItemGroup>
-      <Analyzer Remove="@(_MVVMToolkitAnalyzer)"/>
+      <Analyzer Remove="@(MVVMToolkitAnalyzer)"/>
     </ItemGroup>
   </Target>
 


CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.targets
```diff
@@ -18,27 +18,27 @@
          the same version) but Microsoft.Build.Tasks.CodeAnalysis.dll is where MSBuild loads the compiler tasks from so if
          someone is getting creative with msbuild tasks/targets this is the "most correct" assembly to check. -->
     <GetAssemblyIdentity AssemblyFiles="$([System.IO.Path]::Combine(`$([System.IO.Path]::GetDirectoryName($(CSharpCoreTargetsPath)))`,`Microsoft.Build.Tasks.CodeAnalysis.dll`))">
-      <Output TaskParameter="Assemblies" ItemName="CurrentCompilerAssemblyIdentity"/>
+      <Output TaskParameter="Assemblies" ItemName="MVVMToolkitCurrentCompilerAssemblyIdentity"/>
```
Nirmal4G, line 21:
```suggestion
      <Output TaskParameter="Assemblies" ItemName="_MVVMToolkit_CurrentCompilerAssemblyIdentity"/>
```

CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.targets
```diff
@@ -18,27 +18,27 @@
...
     </GetAssemblyIdentity>
 
     <PropertyGroup>
 
       <!-- Transform the resulting item from GetAssemblyIdentity into a property representing its assembly version -->
-      <CurrentCompilerVersion>@(CurrentCompilerAssemblyIdentity->'%(Version)')</CurrentCompilerVersion>
+      <MVVMToolkitCurrentCompilerVersion>@(MVVMToolkitCurrentCompilerAssemblyIdentity->'%(Version)')</MVVMToolkitCurrentCompilerVersion>
```
Nirmal4G, line 27:
```suggestion
      <_MVVMToolkit_CurrentCompilerVersion>@(_MVVMToolkit_CurrentCompilerAssemblyIdentity->'%(Version)')</_MVVMToolkit_CurrentCompilerVersion>
```

CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.targets
```diff
@@ -18,27 +18,27 @@
...
 
       <!-- The CurrentCompilerVersionIsNotNewEnough property can now be defined based on the Roslyn assembly version -->
-      <CurrentCompilerVersionIsNotNewEnough Condition="$([MSBuild]::VersionLessThan($(CurrentCompilerVersion), 4.0))">true</CurrentCompilerVersionIsNotNewEnough>
+      <MVVMToolkitCurrentCompilerVersionIsNotNewEnough Condition="$([MSBuild]::VersionLessThan($(MVVMToolkitCurrentCompilerVersion), 4.0))">true</MVVMToolkitCurrentCompilerVersionIsNotNewEnough>
```
Nirmal4G, line 30:
```suggestion
      <_MVVMToolkit_CurrentCompilerVersionIsNotNewEnough Condition="$([MSBuild]::VersionLessThan($(_MVVMToolkit_CurrentCompilerVersion), 4.0))">true</_MVVMToolkit_CurrentCompilerVersionIsNotNewEnough>
```

CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.targets
```diff
@@ -18,27 +18,27 @@
...
     </PropertyGroup>
 
     <!-- If the Roslyn version is < 4.0, disable the source generators -->
-    <ItemGroup Condition ="'$(CurrentCompilerVersionIsNotNewEnough)' == 'true'">
+    <ItemGroup Condition ="'$(MVVMToolkitCurrentCompilerVersionIsNotNewEnough)' == 'true'">
```
Nirmal4G, line 34:
```suggestion
    <ItemGroup Condition ="'$(_MVVMToolkit_CurrentCompilerVersionIsNotNewEnough)' == 'true'">
```

CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.targets
```diff
@@ -18,27 +18,27 @@
...
       <Analyzer Remove="@(_MVVMToolkitAnalyzer)"/>
     </ItemGroup>
 
     <!-- If the source generators are disabled, also emit a warning. This would've been produced by MSBuild itself as well, but
          emitting this manually lets us customize the message to inform developers as to why exactly the generators have been
          disabled, and that the rest of the MVVM Toolkit will still keep working as intended, just without additional features. -->
-    <Warning Condition ="'$(CurrentCompilerVersionIsNotNewEnough)' == 'true'" Text="The MVVM Toolkit source generators have been disabled on the current configuration, as they need Roslyn 4.x in order to work. The MVVM Toolkit will work just fine, but features relying on the source generators will not be available."/>
+    <Warning Condition ="'$(MVVMToolkitCurrentCompilerVersionIsNotNewEnough)' == 'true'" Text="The MVVM Toolkit source generators have been disabled on the current configuration, as they need Roslyn 4.x in order to work. The MVVM Toolkit will work just fine, but features relying on the source generators will not be available."/>
```
Nirmal4G, line 41:
```suggestion
    <Warning Condition ="'$(_MVVMToolkit_CurrentCompilerVersionIsNotNewEnough)' == 'true'" Text="The MVVM Toolkit source generators have been disabled on the current configuration, as they need Roslyn 4.x in order to work. The MVVM Toolkit will work just fine, but features relying on the source generators will not be available."/>
```

Nirmal4G (changes requested):
Seems like these properties are not meant to be public. So, you can use a `_` prefix to denote that. But even `_CurrentCompilerVersion` sometimes could be used elsewhere as the name is too generic. Hence, my suggestions!

Either `_` or `_MVVMToolkit_` is fine. But if you think they don't need to be unique then select the former.

Nirmal4G:
There's no official docs on this as far as I know. It's just a convention (_doesn't actually make your property internal/private_) to make sure we don't accidentally use potential MSBuild properties that could be used by users.

You can refer to other props and targets shipped by M$FT and you can see this pattern even in Common props and targets.

Nirmal4G:
There's no official docs on this as far as I know. It's just a convention (_doesn't actually make your property internal/private_) to make sure we don't accidentally use potential MSBuild properties that could be used by users.

You can refer to other props and targets shipped by M$FT and you can see this pattern even in Common props and targets.

Sergio0694:
I see. In that case I don't think we need that, just using "MVVMToolkit" as prefix should be enough. This has been the original suggestion from Sam as well (from the Roslyn team), and what he's been doing as well in the [Antlr4cs project](https://github.com/tunnelvisionlabs/antlr4cs/blob/master/runtime/CSharp/Antlr4BuildTasks/Antlr4.CodeGenerator.targets) too.

Nirmal4G:
Okay, he doesn't seem to use the prefix even for properties that shouldn't be available to user. I'm okay with that. It looks neat. But you do have targets name prefixed. And it's inconsistent with elsewhere, not that I'm complaining but yeah...😅

Sergio0694:
That's a fair point, I think Sam must've also changed his mind on this as he was the one that used "_" in the first place.
I've updated the file in https://github.com/CommunityToolkit/dotnet/pull/321/commits/86aaa9953dbd20ad33a47bba8f76e3cb34af0346 and removed all "_" prefixes, now we just use "MVVMToolkit___" everywhere 🙂

Nirmal4G (approved):
Nice... Looks good!

PR #359, opened 2022-08-02
Replace linker suppressions with attributes

Sergio0694:
As mentioned by @vitek-karas here (https://github.com/dotnet/linker/issues/2938#issuecomment-1202983351), using `#pragma warning disable` for linker annotations is incorrect, as it won't prevent those warnings from actually being emitted. This PR removes the few the MVVM Toolkit had, and just adds the proper annotations to remove the warnings.

cc. @Kant2002

## PR Checklist

### DIFF
--- a/CommunityToolkit.Mvvm/ComponentModel/ObservableValidator.cs
+++ b/CommunityToolkit.Mvvm/ComponentModel/ObservableValidator.cs
@@ -545,9 +545,9 @@ public abstract class ObservableValidator : ObservableObject, INotifyDataErrorIn
         "Additionally, due to the usage of validation APIs, the type of the current instance cannot be statically discovered.")]
     protected void ValidateAllProperties()
     {
-#pragma warning disable IL2026
         // Fast path that tries to create a delegate from a generated type-specific method. This
         // is used to make this method more AOT-friendly and faster, as there is no dynamic code.
+        [RequiresUnreferencedCode("The type of the current instance cannot be statically discovered.")]
         static Action<object> GetValidationAction(Type type)
         {
             if (type.Assembly.GetType("CommunityToolkit.Mvvm.ComponentModel.__Internals.__ObservableValidatorExtensions") is Type extensionsType &&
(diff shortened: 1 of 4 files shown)


CommunityToolkit.Mvvm/ComponentModel/ObservableValidator.cs
```diff
@@ -615,7 +614,9 @@ static Action<object> GetValidationActionFallback(Type type)
 
         // Get or compute the cached list of properties to validate. Here we're using a static lambda to ensure the
         // delegate is cached by the C# compiler, see the related issue at https://github.com/dotnet/roslyn/issues/5835.
-        EntityValidatorMap.GetValue(GetType(), static t => GetValidationAction(t))(this);
+        EntityValidatorMap.GetValue(
+            GetType(),
+            [RequiresUnreferencedCode("The type of the current instance cannot be statically discovered.")] static (t) => GetValidationAction(t))(this);
```
teo-tsirpanis, line 619: You can now just write `GetValidationAction`. The linked Roslyn issue has been closed.

CommunityToolkit.Mvvm/Messaging/IMessengerExtensions.cs
```diff
@@ -233,7 +231,7 @@ public static void RegisterAll<TToken>(this IMessenger messenger, object recipie
         // For more info on this, see the related issue at https://github.com/dotnet/roslyn/issues/5835.
         Action<IMessenger, object, TToken> registrationAction = DiscoveredRecipients<TToken>.RegistrationMethods.GetValue(
             recipient.GetType(),
-            static t => LoadRegistrationMethodsForType(t));
+            [RequiresUnreferencedCode("The type of the current instance cannot be statically discovered.")] static (t) => LoadRegistrationMethodsForType(t));
```
teo-tsirpanis, line 234: Same here.

CommunityToolkit.Mvvm/ComponentModel/ObservableValidator.cs
Sergio0694: The issue is closed but the fix is slated for the C# 11 timeslot, I don't think it's there in VS stable just yet, meaning our CI will be using an older version of Roslyn too without that. I would keep that as is for now (especially given it's unrelated from this PR anyway), and we can update this later on when we bump the SDK as well 🙂

PR #398, opened 2022-08-24
Fix for using C# 8.0 projects (e.g. UWP projects stuck on .NET Core 3)

### LINKED ISSUE #376: Toolkit Doesn't Detect UnSupported C# Version or Doesn't work on C# 8
### Describe the bug

Community Toolkit MVVM Doesn't Support C# Version 8 and Does Only Work with C# Version 9 and Above at the time of writing, In this case there's no proper warning or error output that relates to Community Toolkit MVVM on Compilation

### Regression

7.1.2

### Reproducible in sample app?

- [x] This bug can be reproduced in the sample app.

### Steps to reproduce

```text
1. Create Vanilla UWP Project
2. Install Community Toolkit 8.0 using Nuget
3. Compile the App
```

### Expected behavior

Community Toolkit MVVM Should Have Language Barrier to Prevent the Dependency from Installing on C# version 8 since least version of C# that is supported by Community Toolkit MVVM is C# Version 9 at the time of writing or Community Toolkit MVVM Should Support C# version 8

### Screenshots

Screenshot 2022-08-10 1303041

### Visual Studio Version

2022

### Visual Studio Build Number

17.3

### Nuget packages

Community Toolkit 8.0

### Help us help you

Yes, but only if others can assist.

zhuman:
**Closes #376**

This change fixes support for C# language version 8.0 by swapping a `ImplicitObjectCreationExpression` (which uses the C# 9 language feature `Type myField = new();`) to instead use a normal `ObjectCreationExpression` with the type explicitly defined (`Type myField = new Type();`). This enables the source analyzer to work on .NET Core 3.0 projects once again, such as UWP projects.

As part of testing this fix, I found that the source generator tests as written today don't actually validate successful compilation of _any_ supported scenario, and that a number of scenarios that the tests claim to work on C# 7.3 don't actually work anymore. Therefore I've also changed the following:
* Added "unsupported C# language version" diagnostics to generators which produce C# 8.0 language features. This should be nicer for devs rather than just producing unsupported syntax and hitting random C# compiler errors. A future exercise could be to remove these features if it's desirable to support < C# 8.0.
* For each of the `UnsupportedCSharpLanguageVersion_*` tests, I've also added validation that the source code _does_ compile with C# 8.0.
* I've changed all test source compilation to be done using C# 8.0.
* I've added a dedicated source compilation success test case for ObservableProperty/ObservableObject.

## PR Checklist

### DIFF
--- a/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableValidatorValidateAllPropertiesGenerator.cs
+++ b/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableValidatorValidateAllPropertiesGenerator.cs
@@ -9,7 +9,6 @@ using CommunityToolkit.Mvvm.SourceGenerators.Input.Models;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
-using static CommunityToolkit.Mvvm.SourceGenerators.Diagnostics.DiagnosticDescriptors;
 
 namespace CommunityToolkit.Mvvm.SourceGenerators;
 
@@ -39,8 +38,6 @@ public sealed partial class ObservableValidatorValidateAllPropertiesGenerator : 
             .Where(Execute.IsObservableValidator)
             .Select(static (item, _) => Execute.GetInfo(item))
             .WithComparer(ValidationInfo.Comparer.Default);
-
-        context.FilterWithLanguageVersion(ref validationInfo, LanguageVersion.CSharp8, UnsupportedCSharpLanguageVersionError);
 
         // Check whether the header file is needed
         IncrementalValueProvider<bool> isHeaderFileNeeded =
--- a/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableValidatorValidateAllPropertiesGenerator.cs
+++ b/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableValidatorValidateAllPropertiesGenerator.cs
@@ -9,6 +9,7 @@ using CommunityToolkit.Mvvm.SourceGenerators.Input.Models;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static CommunityToolkit.Mvvm.SourceGenerators.Diagnostics.DiagnosticDescriptors;
 
 namespace CommunityToolkit.Mvvm.SourceGenerators;
 
@@ -38,6 +39,8 @@ public sealed partial class ObservableValidatorValidateAllPropertiesGenerator : 
             .Where(Execute.IsObservableValidator)
             .Select(static (item, _) => Execute.GetInfo(item))
             .WithComparer(ValidationInfo.Comparer.Default);
+
+        context.FilterWithLanguageVersion(ref validationInfo, LanguageVersion.CSharp8, UnsupportedCSharpLanguageVersionError);
 
         // Check whether the header file is needed
         IncrementalValueProvider<bool> isHeaderFileNeeded =
--- a/CommunityToolkit.Mvvm.SourceGenerators/Messaging/IMessengerRegisterAllGenerator.cs
+++ b/CommunityToolkit.Mvvm.SourceGenerators/Messaging/IMessengerRegisterAllGenerator.cs
@@ -10,7 +10,6 @@ using CommunityToolkit.Mvvm.SourceGenerators.Input.Models;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
-using static CommunityToolkit.Mvvm.SourceGenerators.Diagnostics.DiagnosticDescriptors;
 
 namespace CommunityToolkit.Mvvm.SourceGenerators;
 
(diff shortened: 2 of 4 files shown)


zhuman:
I discovered that the unit tests for the generators actually never check for compilation errors at all (it just checks for diagnostics coming directly from the generators, it doesn't check for errors in the generated code), so there are actually other cases where newer C# features are being generated and breaking tests. I am looking at how best to solve these.

Nirmal4G (approved):

CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableValidatorValidateAllPropertiesGenerator.cs
```diff
@@ -39,6 +40,8 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
             .Select(static (item, _) => Execute.GetInfo(item))
             .WithComparer(ValidationInfo.Comparer.Default);
 
+        context.FilterWithLanguageVersion(ref validationInfo, LanguageVersion.CSharp8, UnsupportedCSharpLanguageVersionError);
```
Sergio0694, line 43: The ovservable validator doesn't use C# 8 features, no? This seems unnecessary (it's not free).

CommunityToolkit.Mvvm.SourceGenerators/Messaging/IMessengerRegisterAllGenerator.cs
```diff
@@ -47,6 +48,8 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
             .Select(static (item, _) => Execute.GetInfo(item.Type, item.Interfaces))
             .WithComparer(RecipientInfo.Comparer.Default);
 
+        context.FilterWithLanguageVersion(ref recipientInfo, LanguageVersion.CSharp8, UnsupportedCSharpLanguageVersionError);
```
Sergio0694, line 51: Same here, this generator doesn't need the C# 8 gate, as it doesn't use any C# 8 features.
We can assume C# 7.0 as a baseline.

tests/CommunityToolkit.Mvvm.SourceGenerators.UnitTests/Test_SourceGeneratorsDiagnostics.cs
```diff
@@ -304,9 +315,12 @@ public partial class SampleViewModel : ObservableValidator
                 }
             }";
 
-        // This is explicitly allowed in C# < 8.0, as it doesn't use any new features
         VerifyGeneratedDiagnostics<ObservableValidatorValidateAllPropertiesGenerator>(
```
Sergio0694, line 318: Undo this change (see previous comment), this generator doesn't require C# 8

tests/CommunityToolkit.Mvvm.SourceGenerators.UnitTests/Test_SourceGeneratorsDiagnostics.cs
```diff
@@ -1419,17 +1462,50 @@ private void GreetUser(User user)
...
+            new IMessengerRegisterAllGenerator(),
+            new NullabilityAttributesGenerator(),
+            new ObservableObjectGenerator(),
+            new INotifyPropertyChangedGenerator(),
+            new ObservablePropertyGenerator(),
+            new ObservableRecipientGenerator(),
+            new ObservableValidatorValidateAllPropertiesGenerator(),
+            new RelayCommandGenerator()
+        };
+        VerifyGeneratedDiagnostics(CSharpSyntaxTree.ParseText(source, CSharpParseOptions.Default.WithLanguageVersion(LanguageVersion.CSharp8)), generators, new string[] { });
```
Sergio0694, line 1486: This needs to be updated so that only the lang version required by each generator is used, not just C# 8 for all.

Sergio0694 (changes requested):

CommunityToolkit.Mvvm.SourceGenerators/Messaging/IMessengerRegisterAllGenerator.cs
Nirmal4G: If it doesn't need v8, then we should check against v7.3 (_for UWP_) as indicated in the comments of the attached issue.

CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservableValidatorValidateAllPropertiesGenerator.cs
Sergio0694: No, it isn't. As I said, this adds too much overhead. Besides, I already plan to move this out of the generator.
I'd really like for this PR to only fix that one thing so it's nice and simple and we can merge it quickly.

CommunityToolkit.Mvvm.SourceGenerators/Messaging/IMessengerRegisterAllGenerator.cs
Sergio0694: We can do that, but this is not the right place. I can do that as part of #432.

Sergio0694 (approved):
Pushed a couple tweaks, LGTM! :shipit:

Thank you!

crramirez:
Hello,

One question. Have you tested with UWP projects? Because we are stuck on .Net Core 2.2 and C# 7.3, not .Net Core 3.0. We can use C# 8.0 in .net 2.2 but without all the features: https://www.reflectionit.nl/blog/2019/using-c-8-0-in-core-2-x-net-framework-and-uwp-projects

PR #442, opened 2022-09-15
Changed the namespace of ISettingsStorageHelper interface inline with other helpers.

### LINKED ISSUE #403: ISettingsStorageHelper namespace incorrect
### Describe the bug

Not sure how this was missed, but the ISettingsStorageHelper has a different namespace than all the other helpers here:

https://github.com/CommunityToolkit/dotnet/blob/9d1a525d04299e5421cb3e5917825fee2bc11a18/CommunityToolkit.Common/Helpers/ObjectStorage/ISettingsStorageHelper.cs#L5

All the other helpers are `CommunityToolkit.Common.Helpers`...

https://github.com/CommunityToolkit/dotnet/blob/9d1a525d04299e5421cb3e5917825fee2bc11a18/CommunityToolkit.Common/Helpers/ObjectStorage/IFileStorageHelper.cs#L8

Similarly the extension helper to it has a `.Extensions`:

https://github.com/CommunityToolkit/dotnet/blob/9d1a525d04299e5421cb3e5917825fee2bc11a18/CommunityToolkit.Common/Extensions/ISettingsStorageHelperExtensions.cs#L8

Where the other extensions do not and are just part of `CT.Common`.

### Regression

_No response_

### Steps to reproduce

```text
N/A Code Quality
```

### Expected behavior

Consistency in Namespace

### Screenshots

_No response_

### IDE and version

VS 2022

### IDE version

_No response_

### Nuget packages

- [X] CommunityToolkit.Common

### Nuget package version(s)

8.0.0

### Additional context

Not sure if there's other places to audit?

### Help us help you

Yes, but only if others can assist

maroldo26:
Issue link : https://github.com/CommunityToolkit/dotnet/issues/403

**Closes #403**

Changed the namepspace of the interface ISettingsStorageHelper from **`CommunityToolkit.Helpers`** to **`CommunityToolkit.Common.Helpers`**

## PR Checklist

Existing code which refers `ISettingsStorageHelper` using old namespace `CommunityToolkit.Helpers` will break.

### DIFF
--- a/CommunityToolkit.Common/Extensions/ISettingsStorageHelperExtensions.cs
+++ b/CommunityToolkit.Common/Extensions/ISettingsStorageHelperExtensions.cs
@@ -3,7 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections.Generic;
-using CommunityToolkit.Helpers;
+using CommunityToolkit.Common.Helpers;
 
 namespace CommunityToolkit.Common.Extensions;
 
--- a/CommunityToolkit.Common/Helpers/ObjectStorage/ISettingsStorageHelper.cs
+++ b/CommunityToolkit.Common/Helpers/ObjectStorage/ISettingsStorageHelper.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-namespace CommunityToolkit.Helpers;
+namespace CommunityToolkit.Common.Helpers;
 
 /// <summary>
 /// Service interface used to store data using key value pairs.


Nirmal4G (approved):
This is a straight forward change. I like it but nonetheless, a **huge** breaking change yet not many seem to have used it. So, this might be a good candidate for v8.1 preview to see how the community reacts to the change.

@Sergio0694 Thoughts?

Sergio0694:
I mean I'd really just like to remove the whole bunch of typed in this namespace entirely, but that's a different topic. The changes here look good to me, though yeah this is a breaking change. I think some libs in Microsoft Graph are using these helpers, so we might have to double check with them first that this won't be too disruptive. @michael-hawker do you know who we'd need to ask, just to double check or at the very least give them a heads up?

CommunityToolkit.Common/Extensions/ISettingsStorageHelperExtensions.cs
```diff
@@ -3,7 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections.Generic;
-using CommunityToolkit.Helpers;
+using CommunityToolkit.Common.Helpers;
 
 namespace CommunityToolkit.Common.Extensions;
```
michael-hawker, line 8: There was also the question of here if the `Extensions` should be removed here as no other extensions use this namespace, eh?

michael-hawker:
Yeah, this got introduced in the 8.0 update and was really just a copy-paste issue, unfortunate we didn't catch it before hand, so now it's a breaking change. I don't think we're planning a 9.0 any time soon.

Considering the WCT stuff will ship after you do an update, probably good to just push it through. Not sure how best to communicate a breaking change in a minor release though. It's really just a mistake to correct what was intended vs. us actually trying to change behavior, but it's still a contractual break to semantic versioning... 🙁 Probably would have been easier if we did an 8.0.1 soon after release or something we had noticed it right away.

Nirmal4G:
> It's really just a mistake to correct what was intended vs. us actually trying to change behavior...

It probably depends on how many people depend on this directly. I think you can ship a preview to see how many are breaking, even if it's just a few (_since, most don't use previews_), we can mark this for next major instead!

CommunityToolkit.Common/Extensions/ISettingsStorageHelperExtensions.cs
Nirmal4G: Should this be moved to `Helpers` then?

PR #304, opened 2022-06-09
Update EditorConfig and Git files

Nirmal4G:
## Info

Extracted from CommunityToolkit/WindowsCommunityToolkit#4234 as the Common Code projects were separated in this new repository.

## Changes

- Remove BOM from all files.

- Add default Git attributes.
- Update Git ignores to latest.
- Merge duplicate EditorConfig rules.
- Format projects to EditorConfig rules.
- Fix formatting errors in 'Guard.md' file.

## PR Checklist

- [x] Created a `feature` branch in your fork
- [x] Based off latest main branch of toolkit
- [x] PR doesn't include merge commits (**always rebase on top of our main**)
- [x] Header has been added to all new source files (ran ***build/Update-Headers.ps1***)
- [x] Contains **NO** breaking changes
- [x] Code follows all style conventions

## Notes

- **Rebase** merge please.
- When merging, please update the commit title to PR title instead of the default `Merge pull request #xxxx from repo/branch`, and commit message to either PR message or messages of individual commits. The `auto-merge` option does this by default.

PR #450, opened 2022-09-20
Improve Perf for Azure Pipelines build

Nirmal4G:
## Changes

- Use 'windows-latest' image.
- Narrow Test Results path pattern.
- Do Restore separately from Build step.
- Don't do previous steps like restore/build.
- Use 'BuildConfiguration' for 'Release' build.
- Remove tools that are already present in the image.
- Use Multi-Level lookup to use already installed runtime(s).

___This reduces build times by 20%-40% approx.
If we can cache .NET SDKs and NuGet packages,
we could further reduce build times by another 30%!___

## PR Checklist

## Other information

**Rebase or Squash merge the PR and set its message to title and description.**

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,24 +9,16 @@ pr:
 - rel/*
 
 pool:
-  vmImage: windows-2022
+  vmImage: windows-latest
+
+variables:
+  Build.Configuration: Release
 
 jobs:
 - job: BuildBits
+  displayName: Build and Test solution
   timeoutInMinutes: 60
   steps:
-
-  # Install the .NET Core 3.1 SDK
-  - task: UseDotNet@2
-    displayName: Install the .NET Core 3.1 SDK
-    inputs:
-      version: 3.1.x
-
-  # Install the .NET 6 SDK
-  - task: UseDotNet@2
-    displayName: Install the .NET 6 SDK
-    inputs:
-      version: 6.0.x
 
   # Install the .NET 7 SDK
   - task: UseDotNet@2
@@ -34,20 +26,7 @@ jobs:
     inputs:
       version: 7.0.x
       includePreviewVersions: true
-
-  # Install NuGet
-  - task: NuGetToolInstaller@0
-    displayName: Install NuGet 6.0
-    inputs:
-      versionSpec: 6.0.0
-
-  # Install NerdBank GitVersioning
-  - task: DotNetCoreCLI@2
-    displayName: Install NBGV tool
-    inputs:
-      command: custom
-      custom: tool
-      arguments: update -g nbgv
+      performMultiLevelLookup: true
 
   # Set Build Version
   - script: nbgv cloud
@@ -57,42 +36,48 @@ jobs:
   - pwsh: build/Update-Headers.ps1 -Verify
     displayName: Verify headers
 
+  # Restore solution
+  - script: dotnet restore -p:Configuration=$(Build.Configuration)
+    displayName: Restore solution
+
   # Build solution
-  - script: dotnet build -c Release
+  - script: dotnet build --no-restore -c $(Build.Configuration)
     displayName: Build solution
 
-  # Run .NET 6 tests
-  - script: dotnet test -c Release -f net6.0 -l "trx;LogFileName=VSTestResults_net6.0.trx"
+  # Test solution #
+
+  # Run .NET 6 unit tests
+  - script: dotnet test --no-build -c $(Build.Configuration) -f net6.0 -l "trx;LogFileName=VSTestResults_net6.0.trx"
     displayName: Run .NET 6 unit tests
 
-  # Run .NET Core 3.1 tests
-  - script: dotnet test -c Release -f netcoreapp3.1 -l "trx;LogFileName=VSTestResults_netcoreapp3.1.trx"
+  # Run .NET Core 3.1 unit tests
+  - script: dotnet test --no-build -c $(Build.Configuration) -f netcoreapp3.1 -l "trx;LogFileName=VSTestResults_netcoreapp3.1.trx"
     displayName: Run .NET Core 3.1 unit tests
 
-  # Run .NET Framework 4.7.2 tests
-  - script: dotnet test -c Release -f net472 -l "trx;LogFileName=VSTestResults_net472.trx"
+  # Run .NET Framework 4.7.2 unit tests
+  - script: dotnet test --no-build -c $(Build.Configuration) -f net472 -l "trx;LogFileName=VSTestResults_net472.trx"
     displayName: Run .NET Framework 4.7.2 unit tests
 
   # Publish test results
   - task: PublishTestResults@2
     displayName: Publish test results
     inputs:
-      testResultsFormat: 'VSTest'
-      testResultsFiles: '**/VSTestResults*.trx'
+      testResultsFormat: VSTest
+      testResultsFiles: '**/TestResults/VSTestResults*.trx'
     condition: always()
 
-  # Create the NuGet package(s)
-  - script: dotnet pack -c Release
-    displayName: Create NuGet package(s)
+  # Pack solution
+  - script: dotnet pack --no-build -c $(Build.Configuration)
+    displayName: Pack solution
 
-  # Sign package(s)
+  # Sign packages
   - pwsh: build/Sign-Package.ps1
     displayName: Authenticode sign packages
     env:
       SignClientUser: $(SignClientUser)
       SignClientSecret: $(SignClientSecret)
       ArtifactDirectory: bin/nupkg
-    condition: and(succeeded(), not(eq(variables['Build.Reason'], 'PullRequest')), not(eq(variables['SignClientSecret'], '')), not(eq(variables['SignClientUser'], '')))
+    condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'), ne(variables['SignClientUser'], ''), ne(variables['SignClientSecret'], ''))
 
   # Publish build artifacts
   - publish: bin/nupkg
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,6 +13,8 @@ pool:
 
 variables:
   Build.Configuration: Release
+  DotNet.Tools: $(Agent.ToolsDirectory)/dotnet
+  DotNet.Version: 7.0.x
 
 jobs:
 - job: BuildBits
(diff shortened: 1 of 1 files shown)


Nirmal4G:
Azure Pipeline references:

- [Pipeline caching](https://docs.microsoft.com/azure/devops/pipelines/release/caching)
- [Cache NuGet packages](https://docs.microsoft.com/azure/devops/pipelines/artifacts/caching-nuget)

Nirmal4G:
Reference Commit: a1be7d052fed23fa42bfa28dda98dd22e50bf7e8

Nirmal4G:
Reference Commit: 95ab786fac9e4d5befaee6f8ce7d04a802ad0ad5

azure-pipelines.yml
```diff
@@ -9,45 +9,36 @@ pr:
...
     inputs:
-      version: 6.0.x
+      key: 'dotnet | "$(Agent.OS)" | "$(DotNetVersion)"'
+      restoreKeys: |
+        dotnet | "$(Agent.OS)"
+        dotnet
+      path: $(BuildTools)
 
-  # Install the .NET 7 SDK
+  # Install the .NET 7 SDK (Fixes dotnet/roslyn#63780)
```
Sergio0694, line 35: This is not needed to fix that but, it's needed to make the generators on the Roslyn 4.4 target work at all, as that version of the compiler is bundled with the .NET 7 SDK. I think we can drop this comment here to avoid confusion 🙂

Sergio0694 (changes requested):

azure-pipelines.yml
Nirmal4G: I'll update the comment. I thought it was because of the version of the `S.C.I.` library. I used the latest compiler with `PackageReference` mode and the error still occurred but installing .NET 7 SDK on my machine fixed the issue. So, it could be both.

Sergio0694: There's a couple different bugs in Roslyn 4.3, one is that issue with `S.C.I` you linked, and another is one that causes `ForAttributeWithMetadataName` to crash at runtime, which is why I dropped support for 4.3 and just jumped to 4.4. But that's already documented in the .csproj for the source generator, so it's not needed here. We'd still need the .NET 7 SDK anyway because I'll be adding a .NET 7 target to the HighPerformance shortly, so this bit in the pipeline doesn't need any comments, just like there were none when we were installing the .NET 6 one before 👍

Nirmal4G: Then I'll drop the comment.

Nirmal4G:
Almost done. Roslyn compiler supports `file_header_template`. We could use that instead of `Update-Headers.ps1` script. Remove the invocation from the pipeline.

See the changes here: 8c890b0c7ef0ad7ba48b12bdad53b9ef3aa7d0ec. If you are okay with that, I'll either open a new PR or add it here.

azure-pipelines.yml
```diff
@@ -57,42 +50,58 @@ jobs:
   - pwsh: build/Update-Headers.ps1 -Verify
     displayName: Verify headers
 
+  # Cache NuGet Packages across pipeline runs
```
Sergio0694, line 53: Caching NuGet packages is absolutely overkill, plus I really don't want to add thousands of unmaintainable .lock files that could very easily cause all sorts of issues. It's perfectly fine to just restore NuGet packages normally whenever a build is run. Can you please rebase and drop the last commit? Thank you 🙂

Sergio0694 (changes requested):

azure-pipelines.yml
Nirmal4G: I thought so too. I didn't like it either. I'll drop it.

> Caching NuGet packages is absolutely overkill

**NO** It's not overkill. With caching, I could see the build time drop significantly in subsequent builds. It's definitely worth it. But not when it comes to the maintainability of the project. So, to not have multiple lock files, there's a central lock file (_single lock file for entire repo_) feature that is coming to NuGet CPM. We can add caching to the CPM PR when it's available in time.

Sergio0694:
> "fix was already provided the next day."

We can't rely on that unfortunately. Even if the latest VS 17.3.x build is fixed, that still means that anyone using a previous version (including in CIs) would get random crashes from the generators without a clear explanation as to why. It's just safer to skip to 4.4 directly, which is what I've done. People on 17.3 would still just use the previous generator anyway, which still works, it's just a bit less efficient. 17.4 goes GA in November which is when I plan to release 8.1, so that's fine 🙂

azure-pipelines.yml
Sergio0694: Yes, caching reduces the build time, but even without it, we have no issues. The whole CI still just takes a few minutes, so it's just not worth the extra complexity. It's not like we have hundreds of CIs that take hours to run, like the runtime. Even without any kind of caching, a full CI run just takes like 10 minutes, so I really don't see the value here for all that extra complexity. I'm fine with caching the tools since that's self-contained in the YAML script, but I'm not ok with adding .lock files that would need to be kept in sync, it's just not worth it in our case 🙂

Nirmal4G:
What about the file headers change? https://github.com/CommunityToolkit/dotnet/pull/450#issuecomment-1253882346

Sergio0694:
@Nirmal4G I love that change, but we need files without the header to also produce an error, is that the case?

Nirmal4G:
> we need files without the header to also produce an error, is that the case?

Yes, it does produce error and also suggest a fix for both header less and non-matching scenarios.

Sergio0694:
Awesome, then yeah let's do that too (in another PR) 😄

While at it, can we also normalize the whitespace of all files in the solution, given the header will also have them? All files should just be using \n. Right now I'm not sure there's a line terminator set, and I've seen plenty of files using CLRF instead.

Nirmal4G:
> While at it, can we also normalize the whitespace of all files in the solution, given the header will also have them?

All files are already normalized. I already did this in my previous PRs in Windows repo and recently updated Git Files in here #304.

> All files should just be using \n. Right now, I'm not sure there's a line terminator set, and I've seen plenty of files using CLRF instead.

Git always stores text files with LF (\n) terminator. Git by default checks out and normalizes the line-endings to match each platform's native line-ending (Mac: CR, Unix/Linux: LF, Windows: CRLF).

azure-pipelines.yml
```diff
@@ -9,45 +9,36 @@ pr:
...
+        dotnet | "$(Agent.OS)"
+        dotnet
+      path: $(DotNet.Tools)
 
   # Install the .NET 7 SDK
   - task: UseDotNet@2
     displayName: Install the .NET 7 SDK
     inputs:
-      version: 7.0.x
+      version: $(DotNet.Version)
```
Sergio0694, line 39: We're no longer installing the .NET Core 3.1 SDK, is that just included in the VM image?

azure-pipelines.yml
```diff
@@ -9,45 +9,36 @@ pr:
...
       includePreviewVersions: true
-
-  # Install NuGet
-  - task: NuGetToolInstaller@0
-    displayName: Install NuGet 6.0
-    inputs:
-      versionSpec: 6.0.0
-
-  # Install NerdBank GitVersioning
-  - task: DotNetCoreCLI@2
```
Sergio0694, line 45: How come NBGV was removed?

azure-pipelines.yml
```diff
@@ -9,45 +9,36 @@ pr:
 - rel/*
 
 pool:
-  vmImage: windows-2022
+  vmImage: windows-latest
```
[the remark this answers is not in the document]
Nirmal4G: The `latest` tag rarely changes so no issue here. Even if it does, it is mostly the same set of tools except for the VS version and Windows version which we won't strictly depend on. DevOps team also communicates this via huge header in the pipeline console, so you won't miss it.

All in all, it is better to leave it as `latest`. It's not like we're building "mission critical" projects that need to be deterministic across time.

azure-pipelines.yml
Nirmal4G: Yep.

azure-pipelines.yml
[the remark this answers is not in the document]
Arlodotexe: Why change it though? What's the difference between `windows-2022` and `windows-latest`? 🤔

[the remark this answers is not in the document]
Nirmal4G: @Arlodotexe Both `windows-2022` and `windows-latest` will move to new versioned image when it's broadly available. Except `latest` also moves across major version when it's broadly available.

For Example: _When Dev18 and Win12 get released, and generally available, `latest` will move to that with a 3–6-month migration period. Other than that, everything is the same._

Sergio0694 (changes requested):

azure-pipelines.yml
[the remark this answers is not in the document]
Nirmal4G: > I don't really want to see that approach being used here.

What I mean by that is we don't need to nitpick each and everything. And oh, I read that even windows use latest windows to build itself. And the terminal, dotnet, calculator also uses `windows-latest` similar images to build themselves.

Also, this gives us convenience. We don't need to create PR separately in the future. Anyone can check the pipeline build and know when DevOps team is going to update their images. It gives us 6 months head start, more than what we need to update **if** (_and that's huge if_) there are any breaking changes in the image.

**You'll be good for at least until 2024 in using `windows-latest`**. If all this doesn't convince you, I'll revert it.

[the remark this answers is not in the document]
Sergio0694: Alright, yeah that's fair, let's go with that then 🙂

Sergio0694 (approved):
:shipit:

Sergio0694:
@Nirmal4G Getting this CI failure in https://dev.azure.com/dotnet/CommunityToolkit/_build/results?buildId=79569&view=logs&j=4a47a688-b50e-58f6-4d2c-6d1b829fc8e2&t=2bf27cc2-dcab-5f34-f3ae-ecb203c09e81:
```
C:\hostedtoolcache\windows\dotnet\sdk\7.0.100-rc.1.22431.12\NuGet.targets(396,5): error MSB3202: The project file "D:\a\1\s\tests\CommunityToolkit.Mvvm.Roslyn431.UnitTests\CommunityToolkit.Mvvm.Roslyn431.UnitTests.csproj" was not found. [D:\a\1\s\dotnet Community Toolkit.sln]
```
Is this a regression caused by this PR?

Branch is https://github.com/CommunityToolkit/dotnet/tree/dev/roslyn-431.

Nirmal4G:
@Sergio0694 The error says **Project file not found**. I don't think it is related to this PR. Maybe you forgot to commit this `CommunityToolkit.Mvvm.Roslyn431.UnitTests` renamed file?

PR #452, opened 2022-09-22
Use EditorConfig to apply file headers

Nirmal4G:
## Changes

- Use `file_header_template` to specify our header text in EditorConfig so that the supported editors can use it.
- Remove the `build/Update-Headers.ps1` script since the IDE Analyzer Fixer (**IDE0073**) took over its functionality.
- Remove the `build/header.txt` as the text is embedded in EditorConfig file.

## PR Checklist

- [x] Header has been added to all new source files (Use Visual Studio IDE0073 Code Fix)

## Other information

**Rebase or Squash merge the PR and set its message to title and description.**


Nirmal4G:
One issue I found is that, with script, we can customize where and how we check and apply the header text but with Roslyn analyzer fixer, only source files (i.e. `Compile` items) are used. Other files which might contain headers are not scanned.

Also, it seems to only warn/error in IDE and not in console builds. Do we need to provide an updated script that provides the missing features in the analyzer fixer?

CommunityToolkit.Common/Attributes/NotNullIfNotNullAttribute.cs
```diff
@@ -1,6 +1,6 @@
 // Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
+// The .NET Foundation licenses this file under the MIT license.
```
Sergio0694, line 2: Please undo these changes with a rebase. I said it was great to switch to .editorconfig, but I didn't say we should change the actual header. That's the same standard one across repos, please keep it that way. Even if we did want to change it, it shouldn't be done in the same PR.

Small note, since this keeps happening: contributions are welcome and I do appreciate your help, but please stop adding more changes to every PR that we never discussed. This just makes the review process unnecessarily long due to all the back and forth. Thank you 🙂

Sergio0694 (changes requested):

CommunityToolkit.Common/Attributes/NotNullIfNotNullAttribute.cs
Nirmal4G: Sure, Will move it to a new PR post discussion. I've marked it as Draft because of that. I pushed here just to validate it via CI.

Sergio0694:
> "it seems to only warn/error in IDE and not in console builds"

This is not good, we need the behavior to be the same and builds to fail if the header is not correct everywhere.
If that's no longer the case I don't think we can merge this. 🥲

Nirmal4G:
> we need the behavior to be the same and builds to fail if the header is not correct everywhere.

It does error out for both missing header and mismatched header but when I tried to build via `dotnet build` it doesn't seem to emit the error. But the examples and the videos show errors in the console. Am I missing something here?

Sergio0694:
The current behavior was just to add the header to .cs files, yeah. It's fine to omit it in other files. The important thing is that the CI should fail like before if any .cs file has an invalid header, so we do need the `dotnet build` validation in some way.

Nirmal4G:
I'm getting warnings in the IDE but not getting in `dotnet build`. Can you or anyone from the team can confirm if this is the case? If so, this might be regression in .NET SDK. I can file a report there.

Nirmal4G:
I'm getting warnings in the IDE but not getting in `dotnet build`. Can you or anyone from the team confirm if this is the case? If so, this might be regression in .NET SDK. I can file a report there.

Nirmal4G:
**The build successfully failed!** I never thought I would say this with much enthusiasm.

There was a file `{MVVM.SourceGenerators}\Diagnostics\SuppressionDescriptors.cs` which did not get the header added to it since there was no new line or comment line at the beginning of the file, which is what the script checks for and updates the file with our header text.

Now the `IDE0073` throws for CLI builds. There's a MSBuild property `$(EnforceCodeStyleInBuild)` that must be set in-order for the Style Analyzers to kick-in.

Sergio0694 (approved):
This is awesome, love it! Thanks! 😄

Nirmal4G:
@michael-hawker I can't view pipeline console and logs anymore. The page errors out with **`401-Unauthorized`**. I can access other repo's pipeline's public links but not this. Can you please check this?

<details>
<summary>Attached Screenshot</summary>
<br/>
image
</details>

michael-hawker:
@Nirmal4G I'm not aware of any changes here, does the link open in an inprivate tab? The Azure DevOps can be pretty odd sometimes about authentication even though they should all be public links afaik.

Nirmal4G:
@michael-hawker Yep, it seems to open in InPrivate. 🤔 Why does InPrivate works but not regular way? I do clear all browsing data on session end even on regular tabs. Hmm..., Weird Issue!

PR #481, opened 2022-10-26
Add .NET 7 TFM to HighPerformance target, switch to C# 11

Sergio0694:
This PR introduces the following changes:

- Adds the .NET 7 TFM to the HighPerformance package
- Removes the `Span<T>` trick from the ref-like types, and uses `ref` fields instead
- Enables trimming analyzers in the HighPerformance package again
- Enables C# 11 for the whole solution
- Also uses some C# 11 sugar where needed

## PR Checklist


CommunityToolkit.HighPerformance/Ref{T}.cs
```diff
@@ -2,31 +2,22 @@
...
 namespace CommunityToolkit.HighPerformance;
 
 /// <summary>
 /// A <see langword="struct"/> that can store a reference to a value of a specified type.
 /// </summary>
 /// <typeparam name="T">The type of value to reference.</typeparam>
-[RequiresPreviewFeatures(
-    "The Ref<T> type has no compiler support to ensure the lifetime of referenced values is respected, and as such using it incorrectly may lead to GC holes.",
-    Url = "https://github.com/dotnet/runtime/issues/46104")]
 public readonly ref struct Ref<T>
```
teo-tsirpanis, line 15: How about you make this and `ReadOnlyRef<T>` obsolete? I don't think there is a reason for anyone to use it over regular `ref` fields. Especially since you made it exclusive to .NET 7.

Sergio0694: There's still valid use case scenarios, such as:

- Returning refs to refs. Eg. with this you can have an `out Ref<T>` parameter, which is not otherwise expressible
- I plan on adding more APIs (eg. indexers, etc.) to allow working with GC ref arithmetic easier

Arlodotexe (approved):

PR #485, opened 2022-10-27
Rename solution, readme and icon files

Nirmal4G:
### Rename NuGet Icon to `Icon.png`:
  - This is no longer used as a public reference point for NuGet icon URL.
  - Also, Icon URL is deprecated. Hence, it's safe to change.

### Normalize casing for `ReadMe.md`:
  - To match the file already in Windows toolkit repo.

### Rename solution file to `CommunityToolkit.sln`:
  - The `dotnet` seems implied and also doesn't stand-out in the file list because of the lower casing and `d` char.
  - Spaces are a main issue when doing automation (_like using `*.sln` in build scripts and in URLs it adds `%20`_).

### Misc Changes
  - Don't exclude T4 Generated sources in git ignores list.
  - Indent text in `ThirdPartyNotices.txt` with spaces instead.

## PR Checklist

- [x] Created a separate branch (_other than main/master_) in your fork
- [x] Pull Request doesn't include merge commits
- [x] Header has been added to all new source files

## Other information

 - Please **Rebase** merge if possible.
 - Please update the default message option to `Default to pull request title and description` in the Repository Pull Request settings to have a better commit message instead of `Merge pull request #xxxx from repo/branch` generic message.


Nirmal4G:
These changes are separated from #85, which is becoming too big with the recent refactors. So, I'm extracting some of the changes into separate PRs.

Nirmal4G:
Updated in response to #486

.gitignore
```diff
@@ -15,8 +15,8 @@
 [Ll]og/
 [Ll]ogs/
 
-# Generated file from .ttinclude
-**/Generated/*.g.cs
+# T4 Template Generated files
+!**/Generated/*.g.cs
```
Sergio0694, line 19: I'm confused, if they were ignored before then why do they show up normally in the repo?
Eg. here: https://github.com/CommunityToolkit/dotnet/tree/main/CommunityToolkit.Diagnostics/Generated.

.gitignore
```diff
@@ -31,14 +31,21 @@
 # Visual Studio 15+ auto generated files
 Generated\ Files/
 
+# Backup folder for Package Reference Convert tool in Visual Studio 2017
+MigrationBackup/
```
Sergio0694, line 35: Don't think we need this, we really don't care about Visual Studio 2017, it can't build the repo at all anyway. I know this was here before and it was just moved, but we might just delete it entirely at this point.

.gitignore
```diff
@@ -180,7 +184,7 @@ _pkginfo.txt
 
 # Others
 ClientBin/
-~$*
+~*
```
Sergio0694, line 187: What's this?

CommunityToolkit.sln
```diff
@@ -30,7 +30,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Docs", "Docs", "{88C6FFBE-3
 		CODE_OF_CONDUCT.md = CODE_OF_CONDUCT.md
 		Contributing.md = Contributing.md
 		License.md = License.md
-		README.md = README.md
```
Sergio0694, line 33: Please leave README in uppercase, it's pretty standard.

Sergio0694 (changes requested):

.gitignore
Nirmal4G: If you have ignored files already committed, only subsequent changes are ignored and the files would still be present in source control.

.gitignore
Nirmal4G: The entire git ignore is being updated from upstream and not really specific to this repo except some entries. Even if we delete it here, it would still show up whenever there's an update to those files.

.gitignore
Nirmal4G: Again, update from upstream. This specifically is a POSIX file-system specific ignore.

CommunityToolkit.sln
Nirmal4G: The `README` file without a file extension in uppercase is standard but with a file extension like `.md`, `.txt`, It's usually PascalCase or kebab-case! e.g., `ReadMe.txt` or `Code-of-Conduct.md`. ***Not many people know this.***

The reason I didn't update `CODE_OF_CONDUCT.md` since the updates to the file is managed at org-level. But if you're okay with Normalizing cases then I'll update that file-name too!

.gitignore
Sergio0694: > "The entire git ignore is being updated from upstream"

What do you mean? What upstream?

Nirmal4G: We have an internal fork of the attributes and ignores that's gets updated every once in a while by our dev teams. The one we have is more focused than the general ignores template available on GitHub. That is the upstream. And its also where I got the git attributes, editorconfig files.

Nirmal4G:
Updated in response to #507 and #510

Nirmal4G:
## New Changes since last Review

### Rename `build` folder to `eng`:
  - This is a standard build infra directory used in official dotnet projects.
  - Rename to `eng` folder is done early here to get the `build` folder used for common output paths.

This change is extracted early from patch #96. That PR itself is not yet updated as it depends on changes from this patch and patch #85.

PR #520, opened 2022-12-07
Fix StreamExtensions.Read<T> when using buffered streams

### LINKED ISSUE #513: `Stream.Read<T>()` makes incorrect assumptions about `Stream.Read(Span)`'s return value
### Describe the bug

The `public static unsafe T Read<T>(this Stream stream)` extension method makes the assumption that the `Read(Span<byte>)` must read exactly `length` bytes.
If it returns less than that, it throws with the message that the stream did not contain enough data.
https://github.com/CommunityToolkit/dotnet/blob/2b3e604cf66af5bf1b48fde0bf9efaf45c3195bf/src/CommunityToolkit.HighPerformance/Extensions/StreamExtensions.cs#L209-L218

The only time this assumption is correct is when the stream is a `MemoryStream`.
All other `Stream` implementations are free to return fewer bytes than requested and commonly do so.

### Regression

N/A

### Steps to reproduce

Have a stream like
```c#
class CustomStream : Stream
{
    private ReadOnlyMemory<byte> _bufferedBytes;

    public int Read(Span<byte> buffer)
    {
        if (!_bufferedBytes.IsEmpty)
        {
            int toCopy = Math.Min(_bufferedBytes.Length, buffer.Length);
            _bufferedBytes.Span.Slice(0, toCopy).CopyTo(buffer);
            _bufferedBytes = _bufferedBytes.Slice(toCopy);
            return toCopy;
        }

        return GetMoreBytes(buffer);
    }
}
```

And call
```c#
stream.Read<int>();
```

You may hit an exception telling you that the stream ended even though more data is actually available, you just happened to read while the buffer had fewer than 4 bytes in it.

### Expected behavior

The `Read` helper would loop until either receiving `length` bytes, or until the `Read` returned 0 indicating the end of the stream.

### Screenshots

_No response_

### IDE and version

VS 2022 Preview

### IDE version

_No response_

### Nuget packages

- [ ] CommunityToolkit.Common

### Nuget package version(s)

N/A

### Additional context

I'm not actually using this library, just saw the code as someone mentioned it in dotnet/runtime.

### Help us help you

No, just wanted to report this

Sergio0694:
**Closes #513**

## PR Checklist

### DIFF
--- a/src/CommunityToolkit.HighPerformance/Extensions/StreamExtensions.cs
+++ b/src/CommunityToolkit.HighPerformance/Extensions/StreamExtensions.cs
@@ -207,27 +207,45 @@ public static class StreamExtensions
     {
 #if NETSTANDARD2_1_OR_GREATER
         T result = default;
-        int length = sizeof(T);
+        int bytesOffset = 0;
 
-        unsafe
+        // As per Stream.Read's documentation:
+        // "The total number of bytes read into the buffer. This can be less than the number of bytes allocated in the
+        // buffer if that many bytes are not currently available, or zero (0) if the end of the stream has been reached."
+        // Because of this, we have to loop until all requires bytes have been read, and only throw if the return is 0.
+        do
         {
-            if (stream.Read(new Span<byte>(&result, length)) != length)
+            int bytesRead = stream.Read(new Span<byte>((byte*)&result + bytesOffset, sizeof(T) - bytesOffset));
+
+            // A return value of 0 indicates that the end of the stream has been reached
+            if (bytesRead == 0)
             {
                 ThrowInvalidOperationExceptionForEndOfStream();
             }
+
+            bytesOffset += bytesRead;
         }
+        while (bytesOffset < sizeof(T));
 
         return result;
 #else
-        int length = sizeof(T);
-        byte[] buffer = ArrayPool<byte>.Shared.Rent(length);
+        int bytesOffset = 0;
+        byte[] buffer = ArrayPool<byte>.Shared.Rent(sizeof(T));
 
         try
         {
-            if (stream.Read(buffer, 0, length) != length)
+            do
             {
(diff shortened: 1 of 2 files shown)


MihaZupan (approved):

src/CommunityToolkit.HighPerformance/Extensions/StreamExtensions.cs
```diff
@@ -207,27 +207,45 @@ public static unsafe T Read<T>(this Stream stream)
...
 
-        unsafe
+        // As per Stream.Read's documentation:
+        // "The total number of bytes read into the buffer. This can be less than the number of bytes allocated in the
+        // buffer if that many bytes are not currently available, or zero (0) if the end of the stream has been reached."
+        // Because of this, we have to loop until all requires bytes have been read, and only throw if the return is 0.
+        do
         {
-            if (stream.Read(new Span<byte>(&result, length)) != length)
+            int bytesRead = stream.Read(new Span<byte>((byte*)&result + bytesOffset, sizeof(T) - bytesOffset));
```
gfoidl, line 218: For the .NET 7 target `Stream.ReadExactly` could be used. It boils down the be [almost a one-liner](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABATARgLABQGADAAQY4B0AkgPIDchhGAzOTkuVqQMKkDehUsPJsKnAK4A7AM4BDAGYxSAFVIAlGHIAmAHhUA+ABQAXABYBLGaQDKJ2HIC2pGfa2OAlKQDuZmLFVSEFJpRzkpOQBzGG0hEUECESTA2BkJABsTUgBeUm0YBTkMk0ZE5OFXB0dKTR0AUQQ5MBN0gE8jKRhvWwAHcN1gVpMYYyNB4YAqDwAyVOKAGhcLAC8YCAUjFQ9t0vLhDAB2UjnM3ZEAX0JzoA).

PR #531, opened 2022-12-11
Add analyzer for incorrect [INotifyPropertyChanged] and [ObservableObject] use

### LINKED ISSUE #530: Add analyzer to warn when [INotifyPropertyChanged] & co. are used when not needed
### Overview

The `[INotifyPropertyChanged]` and `[ObservableObject]` are meant to support cases where inheriting isn't possible, but it seems this point is easy to spot for a lot of folks, who just use these attributes instead of the base types. We should consider adding an analyzer to flag those cases and emit a warning. Specifically, these cases should produce a warning:
- `[INotifyPropertyChanged]` on a type that doesn't have a base class.
- `[ObservableObject]` on a type that doesn't have a base class.

### Usage example

The following should produce a warning:

```csharp
[ObservableObject]
public partial class MyViewModel
{
}
```

Sergio0694:
**Closes #530**

## PR Checklist

### DIFF
--- a/src/CommunityToolkit.Mvvm.SourceGenerators/Diagnostics/DiagnosticDescriptors.cs
+++ b/src/CommunityToolkit.Mvvm.SourceGenerators/Diagnostics/DiagnosticDescriptors.cs
@@ -134,7 +134,7 @@ internal static class DiagnosticDescriptors
         id: "MVVMTK0008",
         title: "Unsupported C# language version",
         messageFormat: "The source generator features from the MVVM Toolkit require consuming projects to set the C# language version to at least C# 8.0",
-        category: typeof(CSharpParseOptions).FullName,
+        category: typeof(UnsupportedCSharpLanguageVersionAnalyzer).FullName,
         defaultSeverity: DiagnosticSeverity.Error,
         isEnabledByDefault: true,
         description: "The source generator features from the MVVM Toolkit require consuming projects to set the C# language version to at least C# 8.0. Make sure to add <LangVersion>8.0</LangVersion> (or above) to your .csproj file.",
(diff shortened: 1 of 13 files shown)


src/CommunityToolkit.Mvvm.SourceGenerators/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -507,4 +507,40 @@ internal static class DiagnosticDescriptors
...
+    /// </summary>
+    public static readonly DiagnosticDescriptor InheritFromObservableObjectInsteadOfUsingObservableObjectAttributeWarning = new DiagnosticDescriptor(
+        id: "MVVMTK0033",
+        title: "Inherit from ObservableObject instead of using [ObservableObject]",
+        messageFormat: "The type {0} is using the [ObservableObject] attribute while having no base type, and it should instead inherit from ObservableObject",
+        category: typeof(ObservableObjectGenerator).FullName,
+        defaultSeverity: DiagnosticSeverity.Warning,
+        isEnabledByDefault: true,
+        description:
+            "Classes with no base types should prefer inheriting from [ObservableObject] instead of using attributes to generate INotifyPropertyChanged code, as that will " +
```
Arlodotexe, line 543:
```suggestion
            "Classes with no base types should prefer inheriting from ObservableObject instead of using attributes to generate INotifyPropertyChanged code, as that will " +
```

src/CommunityToolkit.Mvvm.SourceGenerators/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -507,4 +507,40 @@ internal static class DiagnosticDescriptors
...
+    /// </summary>
+    public static readonly DiagnosticDescriptor InheritFromObservableObjectInsteadOfUsingINotifyPropertyChangedAttributeWarning = new DiagnosticDescriptor(
+        id: "MVVMTK0032",
+        title: "Inherit from ObservableObject instead of using [INotifyPropertyChanged]",
+        messageFormat: "The type {0} is using the [INotifyPropertyChanged] attribute while having no base type, and it should instead inherit from ObservableObject",
+        category: typeof(INotifyPropertyChangedGenerator).FullName,
+        defaultSeverity: DiagnosticSeverity.Warning,
+        isEnabledByDefault: true,
+        description:
+            "Classes with no base types should prefer inheriting from [ObservableObject] instead of using attributes to generate INotifyPropertyChanged code, as that will " +
```
Arlodotexe, line 525:
```suggestion
            "Classes with no base types should prefer inheriting from ObservableObject instead of using attributes to generate INotifyPropertyChanged code, as that will " +
```

Arlodotexe (approved):
Just a couple typos, otherwise looks good

PR #539, opened 2022-12-14
Skip field reference diagnostics from constructors

Sergio0694:
This PR is a small follow up for #532. Specifically it solves issues such as this:

```csharp
partial class ViewModel
{
    private string name;

    public ViewModel()
    {
        name = ""; // Warning
    }
}
```

The issue being that:

- If you assign the field, the MVVM Toolkit issues diagnostics
- If you assign the property, you get a nullability warning on the field

The fix is to just not emit diagnostics when assigning fields if the assignment is from a constructor of that type.

cc. @Youssef1313 would you mind also taking a look to confirm if this makes sense? Thank you! 😄

## PR Checklist


Youssef1313 (approved):
Makes sense to me.

Sergio0694:
Awesome, thank you for taking a look! 😊

PR #587, opened 2023-01-26
Resolve all needed symbols early in analyzers

### LINKED ISSUE #586: Resolve symbols early in all analyzers
> context.Compilation.GetTypeByMetadataName

Not so good to do this call here. Better:

```csharp
context.RegisterCompilationStartAction(context =>
{
    var notifyDataErrorInfoAttributeSymbol = context.Compilation.GetTypeByMetadataName("CommunityToolkit.Mvvm.ComponentModel.NotifyDataErrorInfoAttribute");
    if (notifyDataErrorInfoAttributeSymbol is null)
    {
        return;
    }

    context.RegisterSymbolAction(context =>
    {
        // Logic goes here.
    }
}
```

_Originally posted by @Youssef1313 in https://github.com/CommunityToolkit/dotnet/pull/581#discussion_r1084083044_

Sergio0694:
**Closes #586**

## PR Checklist

### DIFF
--- a/src/CommunityToolkit.Mvvm.SourceGenerators/Diagnostics/Analyzers/ClassUsingAttributeInsteadOfInheritanceAnalyzer.cs
+++ b/src/CommunityToolkit.Mvvm.SourceGenerators/Diagnostics/Analyzers/ClassUsingAttributeInsteadOfInheritanceAnalyzer.cs
@@ -68,7 +68,7 @@ public sealed class ClassUsingAttributeInsteadOfInheritanceAnalyzer : Diagnostic
             context.RegisterSymbolAction(context =>
             {
                 // We're looking for class declarations that don't have any base type
-                if (context.Symbol is not INamedTypeSymbol { TypeKind: TypeKind.Class, IsRecord: false, IsStatic: false, IsImplicitlyDeclared: false } classSymbol)
+                if (context.Symbol is not INamedTypeSymbol { TypeKind: TypeKind.Class, IsRecord: false, IsStatic: false, IsImplicitlyDeclared: false, BaseType.SpecialType: SpecialType.System_Object } classSymbol)
                 {
                     return;
                 }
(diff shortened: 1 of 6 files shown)


Sergio0694:
@Youssef1313 thank you for the suggestion! And feel free to take a look 😄

src/CommunityToolkit.Mvvm.SourceGenerators/Extensions/CompilationExtensions.cs
```diff
@@ -87,4 +91,37 @@ public static bool HasAccessibleTypeWithMetadataName(this Compilation compilatio
...
+    /// Tries to build a map of <see cref="INamedTypeSymbol"/> instances form the input mapping of names.
+    /// </summary>
+    /// <typeparam name="T">The type of keys for each symbol.</typeparam>
+    /// <param name="compilation">The <see cref="Compilation"/> to consider for analysis.</param>
+    /// <param name="typeNames">The input mapping of <typeparamref name="T"/> keys to fully qualified type names.</param>
+    /// <param name="typeSymbols">The resulting mapping of <typeparamref name="T"/> keys to resolved <see cref="INamedTypeSymbol"/> instances.</param>
+    /// <returns>Whether all requested <see cref="INamedTypeSymbol"/> instances could be resolved.</returns>
+    public static bool TryBuildNamedTypeSymbolMap<T>(
+        this Compilation compilation,
+        IEnumerable<KeyValuePair<T, string>> typeNames,
```
Youssef1313, line 105: - Curious why this is not a dictionary?
- Assuming all analyzers need *all* types in `typeNames`, LGTM.

src/CommunityToolkit.Mvvm.SourceGenerators/Diagnostics/Analyzers/ClassUsingAttributeInsteadOfInheritanceAnalyzer.cs
```diff
@@ -56,27 +57,31 @@ public override void Initialize(AnalysisContext context)
...
-                    // Next, we need to check whether it isn't already inheriting from another type.
-                    if (classSymbol.BaseType is { SpecialType: SpecialType.System_Object })
+                    return;
+                }
+
+                foreach (AttributeData attribute in context.Symbol.GetAttributes())
+                {
+                    // Same logic as in FieldWithOrphanedDependentObservablePropertyAttributesAnalyzer to find target attributes
+                    if (attribute.AttributeClass is { Name: string attributeName } attributeClass &&
+                        typeSymbols.TryGetValue(attributeName, out INamedTypeSymbol? attributeSymbol) &&
```
Youssef1313, line 80: I guess the dictionary lookup by name is unnecessary?
`typeSymbols` could just be an immutable array and then you can do `typeSymbols.Contains(attributeClass)`

Sergio0694: The idea was that this would avoid doing a linear scan on all symbols, as you'd first lookup by name 🤔

src/CommunityToolkit.Mvvm.SourceGenerators/Extensions/CompilationExtensions.cs
Sergio0694: > "Curious why this is not a dictionary?"

No reason, it was just to give callers more flexibility. I could also make this a dictionary, yeah.

> "Assuming all analyzers need all types in typeNames, LGTM."

Yeah, this would be used by each analyzer to only query all potentially used symbols once at compilation start.

PR #588, opened 2023-01-26
Add fixer for ClassUsingAttributeInsteadOfInheritanceAnalyzer

Sergio0694:
This PR adds a code fixer for `ClassUsingAttributeInsteadOfInheritanceAnalyzer`.

## PR Checklist


Sergio0694:
Hey @Youssef1313, would you have any suggestions on how to handle whitespaces in the code fixer? 😅

src/CommunityToolkit.Mvvm.CodeFixers/ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
```diff
@@ -0,0 +1,135 @@
...
+        // Retrieve the property passed by the analyzer
+        if (diagnostic.Properties[ClassUsingAttributeInsteadOfInheritanceAnalyzer.TypeNameKey] is not string typeName ||
+            diagnostic.Properties[ClassUsingAttributeInsteadOfInheritanceAnalyzer.AttributeTypeNameKey] is not string attributeTypeName)
+        {
+            return;
+        }
+
+        SyntaxNode? root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+        foreach (SyntaxNode syntaxNode in root!.FindNode(diagnosticSpan).DescendantNodesAndSelf())
```
Youssef1313, line 56: Note: Roslyn uses a `GetRequiredSyntaxRootAsync` to avoid suppressions or null checks

https://github.com/dotnet/roslyn/blob/13c1bfbdece3f3890655c5944928e3952189ee05/src/Workspaces/SharedUtilitiesAndExtensions/Workspace/Core/Extensions/DocumentExtensions.cs#L60-L67

src/CommunityToolkit.Mvvm.CodeFixers/ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            // Find the first descendant node from the source of the diagnostic that is a class declaration with the target name
```
Youssef1313, line 58: Why do you want to walk through descendants? Is the diagnostic not on the class declaration?

Youssef1313:
> Hey @Youssef1313, would you have any suggestions on how to handle whitespaces in the code fixer? 😅

Trivia isn't trivial 😄 

Given a quick look, I don't see why there is an extra new line. I'll see if I could debug it.

Youssef1313:
@Sergio0694 The identifier name already has a trailing EndOfLineTrivia, so when you add a base type, the new line remains. You'll have to remove the trivia from the identifier name. But when other interfaces exist, the trailing trivia of identifier name will be a space that you want to keep.

Instead of having to handle all that manually, which is hard to get right (for example, you need to keep trailing trivia after identifier name even if there is no existing base list if the user uses K&R braces), you can use `SyntaxGenerator` to handle everything for you.

https://github.com/CommunityToolkit/dotnet/pull/590

src/CommunityToolkit.Mvvm.CodeFixers/ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
Sergio0694: Ah, yeah good catch, fixed! 🙂

src/CommunityToolkit.Mvvm.CodeFixers/ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
Sergio0694: Eh, I'd rather keep the code self contained for now, I think this is guaranteed to be available anyway.

tests/CommunityToolkit.Mvvm.SourceGenerators.Roslyn401.UnitTests/Test_ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
```diff
@@ -0,0 +1,327 @@
...
+    [DataRow("INotifyPropertyChanged", "MVVMTK0032")]
+    [DataRow("ObservableObject", "MVVMTK0033")]
+    public async Task SingleAttributeList_WithOtherInterface(string attributeTypeName, string diagnosticId)
+    {
+        string original = $$"""
+            using System;
+            using CommunityToolkit.Mvvm.ComponentModel;
+
+            // This is some trivia
+            [{{attributeTypeName}}]
```
Youssef1313, line 74: I don't have much context here, but what if the `ObservableObject` or `INotifyPropertyChanged` attribute is placed on a `record`? Is this a valid scenario?

src/CommunityToolkit.Mvvm.CodeFixers/ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
Youssef1313: Yes it is guaranteed to be non-null.

tests/CommunityToolkit.Mvvm.SourceGenerators.Roslyn401.UnitTests/Test_ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
Sergio0694: No, the attributes only support classes (records are immutable so "property changes" don't really apply.

PR #590, opened 2023-01-28
Use SyntaxGenerator and leave trivia handling up to it

Youssef1313:
@Sergio0694 This PR targets your branch of #588

### DIFF
--- a/src/CommunityToolkit.Mvvm.CodeFixers/ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
+++ b/src/CommunityToolkit.Mvvm.CodeFixers/ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
@@ -64,7 +64,7 @@ public sealed class ClassUsingAttributeInsteadOfInheritanceCodeFixer : CodeFixPr
                 context.RegisterCodeFix(
                     CodeAction.Create(
                         title: "Inherit from ObservableObject",
-                        createChangedDocument: token => UpdateReference(context.Document, root, classDeclaration, attributeTypeName, token),
+                        createChangedDocument: token => UpdateReference(context.Document, root, classDeclaration, attributeTypeName),
                         equivalenceKey: "Inherit from ObservableObject"),
                     diagnostic);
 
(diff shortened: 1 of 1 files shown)


src/CommunityToolkit.Mvvm.CodeFixers/ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
```diff
@@ -76,21 +77,17 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
         // some interfaces in the base list, so we just copy them back after ObservableObject.
-        ClassDeclarationSyntax updatedClassDeclaration =
-            classDeclaration.WithBaseList(BaseList(SingletonSeparatedList(
-            (BaseTypeSyntax)SimpleBaseType(IdentifierName("ObservableObject"))))
-            .AddTypes(classDeclaration.BaseList?.Types.ToArray() ?? Array.Empty<BaseTypeSyntax>()));
-
-        AttributeListSyntax? targetAttributeList = null;
-        AttributeSyntax? targetAttribute = null;
+        SyntaxGenerator generator = SyntaxGenerator.GetGenerator(document);
+        ClassDeclarationSyntax updatedClassDeclaration = (ClassDeclarationSyntax)generator.AddBaseType(classDeclaration, IdentifierName("ObservableObject"));
```
Youssef1313, line 90: Note: what if there is no `using` for the containing namespace of `ObservableObject`?

Sergio0694: Don't think we have to, as it's in the same namespace as the attribute being removed. I assume in theory one could have used the attribute by fully qualified type name instead, but not sure it's worth worrying about 🤔

src/CommunityToolkit.Mvvm.CodeFixers/ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
```diff
@@ -101,35 +97,13 @@ private static async Task<Document> UpdateReference(Document document, ClassDecl
                     (identifierName == attributeTypeName || (identifierName + "Attribute") == attributeTypeName))
                 {
                     // We found the attribute to remove and the list to update
-                    targetAttributeList = attributeList;
-                    targetAttribute = attribute;
+                    updatedClassDeclaration = (ClassDeclarationSyntax)generator.RemoveNode(updatedClassDeclaration, attribute);
```
Sergio0694, line 100: I assume this will also take care of removing the whole attribute list if it becomes empty?

Sergio0694:
Oh wow, TIL, this is awesome!! 😄

src/CommunityToolkit.Mvvm.CodeFixers/ClassUsingAttributeInsteadOfInheritanceCodeFixer.cs
Youssef1313: @Sergio0694 Yes. Roslyn implementation of this part:

https://github.com/dotnet/roslyn/blob/13c1bfbdece3f3890655c5944928e3952189ee05/src/Workspaces/CSharp/Portable/CodeGeneration/CSharpSyntaxGenerator.cs#L2994-L3002

Sergio0694 (approved):
LGTM, thank you so much!! 🙌

PR #646, opened 2023-03-14
Add [MemberNotNull] to [ObservableProperty] set accessors

### LINKED ISSUE #645: Add [MemberNotNull] on [ObservableProperty] setter when needed
### Overview

Currently, `[ObservableProperty]` has an issue with nullability annotations. Consider this:

```csharp
[ObservableProperty]
private string name;
```

Even if you set `Name` in the constructor, you'll get a warning saying "name might be null", as the compiler doesn't know that setting `Name` (to a non null value) also means `name` is set to a non null value. We can fix this by generating:

```csharp
public string Name
{
    get => name;

    [MemberNotNull(nameof(name))]
    set => name = value;
}
```

That is, if:
- The field is a non nullable reference type
- Nullable annotations are enabled in the file
- `[MemberNotNull]` is available

Then also add `[MemberNotNull]` in the setter, referring the field.

This way you can set the property in the constructor, and Roslyn will no longer warn that the field might be null and it's not set.

### Usage example

The following should not produce a warning:

```csharp
using CommunityToolkit.Mvvm.ComponentModel;

public partial class MyViewModel : ObservableObject
{
    public MyViewModel()
    {
        Name = "Bob";
    }

    [ObservableProperty]
    private string name;
}
```

### Breaking change?

No

Sergio0694:
**Closes #645**

cc. @Youssef1313 for your consideration if you have spare time once again 😄

## PR Checklist


src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservablePropertyGenerator.Execute.cs
```diff
@@ -668,6 +670,36 @@ private static bool IsOldPropertyValueDirectlyReferenced(IFieldSymbol fieldSymbo
...
+            //    }
+            //
+            //    [ObservableProperty]
+            //    private string name;
+            // }
+            //
+            // The [MemberNotNull] attribute is needed on the setter for the generated Name property so that when Name
+            // is set, the compiler can determine that the name backing field is also being set (to a non null value).
+            return
+                fieldSymbol.Type is { NullableAnnotation: not NullableAnnotation.Annotated, IsReferenceType: true } &&
```
Youssef1313, line 699: I don't know whether generic types are supported in this context. If yes, consider adding a test for a generic type that is constrained to `class`. I'm not sure if this will require additional handling or if `IsReferenceType` will be sufficient.

Sergio0694: Oooh good catch! I'm seeing the warning even without the constraint though:

image

I'm thinking it might make sense to emit the attribute in all cases except if there's a `struct` constraint? 🤔
Including `unmanaged` as well, as a subset of that anyway. What do you think?

Youssef1313: @Sergio0694 maybe exclude `notnull` constraint as well?

Sergio0694: Why? If `T : notnull` it means the type is not nullable, so if you have a `T foo` field you'd definitely want the attribute on it, otherwise you'll also get a warning due to the field potentially being null, no? 🤔

I'm thinking essentially we should forward the attribute whenever the type is not annotated (which also excludes nullable value types) and not explicitly a value type (which covers reference types and unconstrained type parameters without `struct` or `unmanaged`).

Youssef1313: Oh, indeed. You're right.

Sergio0694: I've updated the generators and added a whole bunch of tests in https://github.com/CommunityToolkit/dotnet/pull/646/commits/de885cb48256c7317990183d329c19333d5cfff1, let me know what you think! 😄

src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservablePropertyGenerator.Execute.cs
```diff
@@ -668,6 +675,46 @@ private static bool IsOldPropertyValueDirectlyReferenced(IFieldSymbol fieldSymbo
...
+        /// <param name="isReferenceTypeOrUnconstraindTypeParameter">Whether the property type supports nullability.</param>
+        /// <param name="includeMemberNotNullOnSetAccessor">Whether <see cref="MemberNotNullAttribute"/> should be used on the setter.</param>
+        /// <returns></returns>
+        private static void GetNullabilityInfo(
+            IFieldSymbol fieldSymbol,
+            SemanticModel semanticModel,
+            out bool isReferenceTypeOrUnconstraindTypeParameter,
+            out bool includeMemberNotNullOnSetAccessor)
+        {
+            // We're using IsValueType here and not IsReferenceType to also conver unconstrained type parameter cases.
```
Youssef1313, line 692:
```suggestion
            // We're using IsValueType here and not IsReferenceType to also cover unconstrained type parameter cases.
```

src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservablePropertyGenerator.Execute.cs
```diff
@@ -668,6 +675,46 @@ private static bool IsOldPropertyValueDirectlyReferenced(IFieldSymbol fieldSymbo
...
+            //    }
+            //
+            //    [ObservableProperty]
+            //    private string name;
+            // }
+            //
+            // The [MemberNotNull] attribute is needed on the setter for the generated Name property so that when Name
+            // is set, the compiler can determine that the name backing field is also being set (to a non null value).
+            includeMemberNotNullOnSetAccessor =
+                fieldSymbol.Type is { NullableAnnotation: not NullableAnnotation.Annotated, IsValueType: false } &&
```
Youssef1313, line 714: Stylistic suggestion. `IsValueType` was already checked above, so reused `isReferenceTypeOrUnconstraindTypeParameter`. Then the pattern ended up with a single property, so checked it directly. You can definitely ignore this if you like the original code more.

```suggestion
            includeMemberNotNullOnSetAccessor =
                isReferenceTypeOrUnconstraindTypeParameter &&
                fieldSymbol.Type.NullableAnnotation != NullableAnnotation.Annotated &&
```

Youssef1313 (approved):

src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/Models/PropertyInfo.cs
```diff
@@ -30,5 +31,6 @@ internal sealed record PropertyInfo(
     bool NotifyPropertyChangedRecipients,
     bool NotifyDataErrorInfo,
     bool IsOldPropertyValueDirectlyReferenced,
-    bool IsReferenceType,
+    bool IsReferenceTypeOrUnconstraindTypeParameter,
+    bool IncludeMemberNotNullOnSetAccessor,
```
jkoritzinsky, line 35: With this many `bool` properties, I'd recommend switching to a bitmask personally. Especially if you ever plan to add unit testing that will construct this with specific values (instead of always passing in well-named variables as you do today).

src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservablePropertyGenerator.Execute.cs
Sergio0694: Makes sense! I've integrated the changes in the new commit with the other tweaks too. Thank you! 🙂

PR #650, opened 2023-03-18
Fix Guard IsWhiteSpace and IsNotWhiteSpace Methods

### LINKED ISSUE #441: Guard.IsNotWhiteSpace gives exception on null value
Describe the bug
Null check done in Guard.IsNotWhiteSpace while Guard.IsNotNullOrWhiteSpace is also available.
Code of Guard.IsNotWhiteSpace has the same implementation as Guard.IsNotNullOrWhiteSpace. 
null does not feel equal to whitespace. 

Regression
No response

Steps to reproduce
Call Guard.IsNotWhiteSpace(null), this gives a System.ArgumentException.

Screenshots
No response

IDE and version
VS 2017

IDE version
No response

Nuget packages
- [ ] CommunityToolkit.Common
- [X] CommunityToolkit.Diagnostics

Nuget package version(s)
8.0.0

Additional context
No response

Help us help you
No, just wanted to report this

GabrieleMessina:
**Closes #441**

Fixed Guard IsWhiteSpace and IsNotWhiteSpace methods adding a check for null before calling string.IsNullOrWhiteSpace. 
Added some tests for both methods.

## PR Checklist

- [ ] New component

## Other information

I don't know if this can be considered a breaking change.

### DIFF
--- a/src/CommunityToolkit.Diagnostics/Guard.String.cs
+++ b/src/CommunityToolkit.Diagnostics/Guard.String.cs
@@ -118,15 +118,15 @@ partial class Guard
     }
 
     /// <summary>
-    /// Asserts that the input <see cref="string"/> instance must be whitespace.
+    /// Asserts that the input <see cref="string"/> instance must consists exclusively of white-space characters.
     /// </summary>
     /// <param name="text">The input <see cref="string"/> instance to test.</param>
     /// <param name="name">The name of the input parameter being tested.</param>
-    /// <exception cref="ArgumentException">Thrown if <paramref name="text"/> is neither <see langword="null"/> nor whitespace.</exception>
+    /// <exception cref="ArgumentException">Thrown if <paramref name="text"/> has some not white-space characters.</exception>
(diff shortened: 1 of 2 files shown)


cremor:
In my opinion those methods are still confusing with this PR. Why does "white space" include an empty string, but doesn't include null? Neither are white-space characters. Just from the method name I'd assume that an empty string would fail the `Guard.IsWhiteSpace` call.

But I can see how the method name is similar to `String.IsNullOrWhiteSpace`, which also includes empty strings. But that method also explicitly mentions empty strings in its documentation.

So, at least I'd suggest to change the documentation of those methods to also mention empty strings.
For even better distinction I would suggest to only allow white-space characters in `Guard.IsWhiteSpace` and add additional methods for `Guard.IsEmptyOrWhiteSpace`.

GabrieleMessina:
@cremor I think you have a point. I'll change the code as soon as I can. 
I'm not so sure, though, about the new `Guard.IsEmptyOrWhiteSpace` method because it can be achieved using a combination of the other Guard methods and I don't think it'll be widely used.

PR #701, opened 2023-05-16
Improve ImmutableArrayBuilder<T> type in MVVM Toolkit generators

Sergio0694:
## Overview

This PR includes a couple tweaks to `ImmutableArrayBuilder<T>`:
- Drops the `ObjectPool<T>` use, and just uses `ArrayPool<T>`. This means we can now also benefit from the automatic trimming that `ArrayPool<T>` implements if there is high GC pressure. Also, we're no longer reinventing the wheel here.
- Make the type a `ref struct`, since we're never meant to pass this outside or box it.
- Add ref scope annotations where needed

## PR Checklist


src/CommunityToolkit.Mvvm.SourceGenerators/Helpers/ImmutableArrayBuilder{T}.cs
```diff
@@ -15,13 +17,8 @@ namespace CommunityToolkit.Mvvm.SourceGenerators.Helpers;
 /// A helper type to build sequences of values with pooled buffers.
 /// </summary>
 /// <typeparam name="T">The type of items to create sequences for.</typeparam>
-internal struct ImmutableArrayBuilder<T> : IDisposable
+internal ref struct ImmutableArrayBuilder<T>
```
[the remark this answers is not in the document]
Sergio0694: Can't implement interfaces on `ref struct` types. But Roslyn has special support for them via duck-typing, so as long as they expose a `public void Dispose()` method, you can still use such types in `using` statements/expressions 🙂

src/CommunityToolkit.Mvvm.SourceGenerators/Helpers/ImmutableArrayBuilder{T}.cs
```diff
@@ -96,30 +94,25 @@ public int Count
...
     /// A class handling the actual buffer writing.
     /// </summary>
-    private sealed class Writer
+    private sealed class Writer : IDisposable
     {
         /// <summary>
         /// The underlying <typeparamref name="T"/> array.
         /// </summary>
-        private T[] array;
+        private T?[]? array;
```
Youssef1313, line 115: The array is assigned to non-null in constructor. Anything I'm missing?

Youssef1313: Oh I see, it's null-ed in Dispose.

Youssef1313 (approved):
LGTM

PR #704, opened 2023-05-16
Check for cancellation more often in generators

### LINKED ISSUE #703: Source generators should check for cancellation more often
### Overview

Currently, the MVVM Toolkit source generators are propagating the cancellation token in each callback to all calls that need one, but they're not checking for cancellation themselves that much. This can increase overhead when concurrent invocations are triggered, as previous runs of the generators will take longer before they get to a point where they can detect they've been cancelled. We should update them to add more cancellation checks in places that make sense.

Sergio0694:
**Closes #703**

image

## PR Checklist


Sergio0694:
This seems pretty straightfoward, but anything in particular I'm missing to make this even better by any chance? 😄
cc. @Youssef1313

Youssef1313 (approved):
Too much `ThrowIfCancellationRequested` 😄 , but LGTM

PR #719, opened 2023-06-20
Remove unnecessary temporary array allocation

Sergio0694:
## Overview

This PR adds an API to enumerate items in an array builder without allocating a whole array.
It also uses this API when creating a `BlockSyntax`, to avoid wasting an array just to enumerate items.
The returned enumerable is readonly, ensuring mutations must go through the actual builder.

## PR Checklist

### DIFF
--- a/src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservablePropertyGenerator.Execute.cs
+++ b/src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservablePropertyGenerator.Execute.cs
@@ -961,7 +961,7 @@ partial class ObservablePropertyGenerator
                         .AddArgumentListArguments(
                             Argument(fieldExpression),
                             Argument(IdentifierName("value")))),
-                    Block(setterStatements.ToArray()));
+                    Block(setterStatements.AsEnumerable()));
 
             // Prepare the forwarded attributes, if any
             ImmutableArray<AttributeListSyntax> forwardedAttributes =
--- a/src/CommunityToolkit.Mvvm.SourceGenerators/Helpers/ImmutableArrayBuilder{T}.cs
+++ b/src/CommunityToolkit.Mvvm.SourceGenerators/Helpers/ImmutableArrayBuilder{T}.cs
@@ -7,6 +7,8 @@
 
 using System;
 using System.Buffers;
+using System.Collections;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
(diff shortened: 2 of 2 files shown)


Sergio0694:
Hey @Youssef1313, want to take a quick lock when you have a minute? This one is very smol 😄

src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservablePropertyGenerator.Execute.cs
```diff
@@ -961,7 +961,7 @@ public static MemberDeclarationSyntax GetPropertySyntax(PropertyInfo propertyInf
                         .AddArgumentListArguments(
                             Argument(fieldExpression),
                             Argument(IdentifierName("value")))),
-                    Block(setterStatements.ToArray()));
+                    Block(setterStatements.AsEnumerable()));
```
Youssef1313, line 964: Did the allocation show up in traces?

src/CommunityToolkit.Mvvm.SourceGenerators/Helpers/ImmutableArrayBuilder{T}.cs
```diff
@@ -173,6 +190,48 @@ public void Dispose()
...
+
+        /// <inheritdoc/>
+        IEnumerator<T> IEnumerable<T>.GetEnumerator()
+        {
+            T?[] array = this.array!;
+            int length = this.index;
+
+            for (int i = 0; i < length; i++)
+            {
+                yield return array[i]!;
```
Youssef1313, line 219: Wouldn't this allocate a state machine? Why is that better than allocating the array?

Sergio0694: Well, the array would just be on top of that. As in, if we create the array and pass that to Roslyn, it will then enumerate it anyway (as an `IEnumerable<T>`), so we'll still allocate an enumerator for that here. So this way instead we can at least skip the array entirely (which is just intermediate and wasted) and just enumerate directly 🙂
This also matches what you can do with the normal `ImmutableArray<T>.Builder` type.

src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservablePropertyGenerator.Execute.cs
Sergio0694: Not really, that wasted allocation was just bothering me 😄
And there's plenty more like that in other generators.

src/CommunityToolkit.Mvvm.SourceGenerators/Helpers/ImmutableArrayBuilder{T}.cs
Youssef1313: Makes sense!

src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/ObservablePropertyGenerator.Execute.cs
Youssef1313: Yup likely the impact is low here, but good to remove an unnecessary allocation anyway.

Sergio0694:
@Youssef1313 I assume you have no other concerns with this one, so good to go? 😄

Youssef1313 (approved):
Yup. Looks good!

PR #769, opened 2023-10-10
Test for duplicate interface implementation instead of declaration

### LINKED ISSUE #620: Enable other classes that already implement the `INotifyPropertyChanged` to leverage Source Generator Attributes
### Overview

For instance, I already have a class I'm using which inherits from `INotifyPropertyChanged`. In my scenario, it's `ObservableCollection`, so I can't modify it as it's in the BCL.

```cs
public class ObservableCollection<T> : Collection<T>, INotifyCollectionChanged, INotifyPropertyChanged
```

It already has an `OnPropertyChanged` helper and `PropertyChanged` event:

```cs
        protected virtual event PropertyChangedEventHandler? PropertyChanged;

        protected virtual void OnPropertyChanged(PropertyChangedEventArgs e)
        {
            PropertyChanged?.Invoke(this, e);
        }
```

It would be great if I could still use some or all of the MVVM attributes for source generators like `[ObservableProperty]` when inheriting from these classes in my own class.

### API breakdown

N/A uses existing attributes.

### Usage example

For instance, I have a class already inheriting from this:

```cs
public partial class MyClass : ObservableCollection<T>
{
     // Since my parent class already inherits from INotifyPropertyChanged (and already has a OnPropertyChanged method), just use those...
     [ObservableProperty]
     private bool _isModified;
}
```

Worst case, let me have to re-specify that I want whatever's needed to polyfill, either with `INotifyPropertyChanged` or `ObservableObject`:

```cs
[INotifyPropertyChanged]
public partial class MyClass : ObservableCollection<T>
{
     [ObservableProperty]
     private bool _isModified;

     partial OnIsModifiedChanged(bool value)
     {
         // Yay, I could also do this now...
     }
}
```

Basically the existing generator in this case would spit out all the same stuff except what's already implemented by the type (in the case of `ObservableCollection` it'd be the following:

```cs
    partial class MyClass : global::System.ComponentModel.INotifyPropertyChanged
    {
        //// The Event and the event args OnPropertyChanged are already implemented, so skip adding them

        /// <summary>
        /// Raises the <see cref = "PropertyChanged"/> event.
        /// </summary>
        /// <param name = "propertyName">(optional) The name of the property that changed.</param>
        [global::System.CodeDom.Compiler.GeneratedCode("CommunityToolkit.Mvvm.SourceGenerators.INotifyPropertyChangedGenerator", "8.1.0.0")]
        [global::System.Diagnostics.DebuggerNonUserCode]
        [global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
        protected void OnPropertyChanged([global::System.Runtime.CompilerServices.CallerMemberName] string? propertyName = null)
        {
            OnPropertyChanged(new global::System.ComponentModel.PropertyChangedEventArgs(propertyName));
        }

        //// All the SetProperty helpers, TaskNotifier, etc...
```

If there's any conflict for some particular type, then raise an analytic warning/error.

### Breaking change?

No

### Alternatives

Can't use MVVM Toolkit generators or helpers? 😥

### Additional context

Don't know if I know enough to help here. Had a hard time trying to understand how the existing code was generated with the new fanciness...

### Help us help you

No, just wanted to propose this

### LINKED ISSUE #805: Generating of `[ObservableObject]` fails if interface is declared but not implemented
### Describe the bug

If a class implements an interface that itself inherits `INotifyPropertyChanged` or `INotifyPropertyChanging` but does not implement the interface methods, code-generation aborts.

Current tests only test diagnostics for an _implementation_ of `INotifyPropertyChanged` / `INotifyPropertyChanging`:
https://github.com/CommunityToolkit/dotnet/blob/7b53ae23dfc6a7fb12d0fc058b89b6e948f48448/tests/CommunityToolkit.Mvvm.SourceGenerators.UnitTests/Test_SourceGeneratorsDiagnostics.cs#L33-L37

### Regression

_No response_

### Steps to reproduce

```cs
interface IStuff : INotifyPropertyChanged {
   // ...
}

[ObservableObject]
class B : BaseClass, IStuff {
   // No implementation of INotifyPropertyChanged 
   // source generation
}
```

### Expected behavior

The implementation of `INotifyPropertyChanged` should be generated.

### Screenshots

_No response_

### IDE and version

VS 2022

### IDE version

_No response_

### Nuget packages

- [X] CommunityToolkit.Mvvm (aka MVVM Toolkit)

### Nuget package version(s)

8.2.2

### Additional context

Similar to #620 but slightly different scenario

### Help us help you

ShortDevelopment:
**Closes #805**

## PR Checklist

### DIFF
--- a/src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/INotifyPropertyChangedGenerator.cs
+++ b/src/CommunityToolkit.Mvvm.SourceGenerators/ComponentModel/INotifyPropertyChangedGenerator.cs
@@ -33,14 +33,12 @@ public sealed class INotifyPropertyChangedGenerator : TransitiveMembersGenerator
     {
         diagnostics = ImmutableArray<DiagnosticInfo>.Empty;
 
-        INotifyPropertyChangedInfo? info = null;
-
         // Check if the type already implements INotifyPropertyChanged
-        if (typeSymbol.AllInterfaces.Any(i => i.HasFullyQualifiedMetadataName("System.ComponentModel.INotifyPropertyChanged")))
+        if (typeSymbol.ImplementsInterfaceMember("System.ComponentModel.INotifyPropertyChanged"))
         {
             diagnostics = ImmutableArray.Create(DiagnosticInfo.Create(DuplicateINotifyPropertyChangedInterfaceForINotifyPropertyChangedAttributeError, typeSymbol, typeSymbol));
 
-            goto End;
+            return null;
         }
 
         // Check if the type uses [INotifyPropertyChanged] or [ObservableObject] already (in the type hierarchy too)
@@ -49,15 +47,12 @@ public sealed class INotifyPropertyChangedGenerator : TransitiveMembersGenerator
         {
             diagnostics = ImmutableArray.Create(DiagnosticInfo.Create(InvalidAttributeCombinationForINotifyPropertyChangedAttributeError, typeSymbol, typeSymbol));
 
-            goto End;
+            return null;
         }
 
         bool includeAdditionalHelperMethods = attributeData.GetNamedArgument("IncludeAdditionalHelperMethods", true);
 
-        info = new INotifyPropertyChangedInfo(includeAdditionalHelperMethods);
-
-        End:
-        return info;
+        return new INotifyPropertyChangedInfo(includeAdditionalHelperMethods);
     }
 
     /// <inheritdoc/>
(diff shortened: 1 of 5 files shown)


ShortDevelopment:
@dotnet-policy-service agree

ShortDevelopment:
@Sergio0694 Sorry, for pinging you, but I'm kind of blocked by this.
Could you please have a look at this PR?

PR #823, opened 2024-01-09
Fast unboxing of Box<T> value without unnecessary type checking 

### LINKED ISSUE #822: Fast unboxing of Box<T> value without unnecessary type checking
### Overview

Box<T> uses a slower unboxing process, making sure that the instance can be converted to type T. However, since the instance creation guarantees that the correct type of value is enclosed, we can make the unboxing methods quicker by eliminating the unnecessary type checking.

### API breakdown

```csharp
public sealed class Box<T>
    where T : struct
{
+    internal readonly T value; // used for fast unboxing

    public static implicit operator T(Box<T> box)
        => box.value; // Fast unboxing without type checking
}

public static class BoxExtensions
{
    public static ref T GetReference<T>(this Box<T> box) where T : struct
        => ref Unsafe.AsRef(box.value); // Fast unboxing without type checking
}
```

### Usage example

The same as before but faster.

### Breaking change?

No

### Alternatives

I don't know.

### Additional context

_No response_

### Help us help you

AlexRadch:
Closes #822

## PR Checklist

- [ ] New component
- [ ] Code follows all style conventions

### DIFF
--- a/src/CommunityToolkit.HighPerformance/Box{T}.cs
+++ b/src/CommunityToolkit.HighPerformance/Box{T}.cs
@@ -125,7 +125,10 @@ public sealed class Box<T>
     /// <param name="box">The input <see cref="Box{T}"/> instance.</param>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static implicit operator T(Box<T> box)
-        => box.value;
+    {
+        return box.value;
+    }
+        
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


Sergio0694:
The code being introduced relies on undefined behavior and is not safe. The whole type really is doing that too, which is why I plan on removing it altogether. There is no safe way of unboxing a boxed value bypassing type checks as of now, and relying on runtime internals is not the right approach. I also somehow missed marking this type as obsolete in the past 8.x release 😅

PR #825, opened 2024-01-10
Box<T>.GetValue()

### LINKED ISSUE #824: Add Box<T>.GetValue() extension method
### Overview

Now to get the boxed value, we need to use an implicit cast to the boxed type. The compiler cannot always make such an implicit cast, especially for generic methods. At first, the compiler tries to pass a Box<T> instance to a generic method and does not perform implicit type casting. In this case, developers are forced to call an explicit instance cast to the boxed type. For example `Assert.AreEqual(value, (boxedType)box)`. I believe that it will be more convenient for developers to use the **`GetValue()`** extension method to get the boxed value, rather than calling an explicit cast to the boxed type.

Using the `GetReference()` method is less preferable because this method is unsafe and allows the boxed value to be modified. The `GetReference()` method is intended primarily for rare scenarios where changing a boxed value without creating a new instance can improve code performance and the developer understands why he is using this method.

So I suggest adding an extension method `Box<T>.GetValue()` to safely get the boxed value. This method will be safe, developer-friendly, and compiler-friendly in generic methods.

### API breakdown

```csharp
public static class BoxExtensions
{
    public static T GetValue<T>(this Box<T> box) where T : struct => (T)box;
}
```

### Usage example

`Assert.AreEqual(value, box.GetValue());`

### Breaking change?

No

### Alternatives

Use the dangerous `GetReference()` method.

### Additional context

_No response_

### Help us help you

AlexRadch:
**Closes #824**

## PR Checklist

- [ ] New component

### DIFF
--- a/src/CommunityToolkit.HighPerformance/Box{T}.cs
+++ b/src/CommunityToolkit.HighPerformance/Box{T}.cs
@@ -218,4 +218,35 @@ public static class BoxExtensions
         // this doesn't really matter in this case anyway.
         return ref Unsafe.Unbox<T>(box);
     }
+
+    /// <summary>
+    /// Gets a <typeparamref name="T"/> value from a <see cref="Box{T}"/> instance.
+    /// </summary>
+    /// <typeparam name="T">The type of value to retrieve.</typeparam>
+    /// <param name="box">The input <see cref="Box{T}"/> instance.</param>
+    /// <returns>The <typeparamref name="T"/> boxed value within <paramref name="box"/>.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static T GetValue<T>(this Box<T> box)
+        where T : struct
+    {
+        // The reason why this method is an extension and is not part of
+        // the Box<T> type itself is that Box<T> is really just a mask
+        // used over object references, but it is never actually instantiated.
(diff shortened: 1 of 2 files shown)


src/CommunityToolkit.HighPerformance/Box{T}.cs
```diff
@@ -218,4 +218,35 @@ public static class BoxExtensions
...
     }
+
+    /// <summary>
+    /// Gets a <typeparamref name="T"/> value from a <see cref="Box{T}"/> instance.
+    /// </summary>
+    /// <typeparam name="T">The type of value to retrieve.</typeparam>
+    /// <param name="box">The input <see cref="Box{T}"/> instance.</param>
+    /// <returns>The <typeparamref name="T"/> boxed value within <paramref name="box"/>.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static T GetValue<T>(this Box<T> box)
```
Sergio0694, line 229: As a side point to the fact we're not planning on adding any new APIs to the `Box<T>` type, keep in mind that this method is also not really doing anything anyway, as there's already the implicit operator that's doing this. So this would really just be a slightly different syntax for doing what is already possible via just implicit conversion (or cast).

PR #966, opened 2024-10-25
Add support for partial properties for [ObservableProperty]

### LINKED ISSUE #555: Add support for [ObservableProperty] on partial properties
### Overview

This issue tracks adding support for partial properties to the `[ObservableProperty]` generator. This is blocked on https://github.com/dotnet/csharplang/issues/6420, so support for this will have to wait for that first (hopefully in the C# 12 timeframe).

### API breakdown

The changes are pretty straightforward: properties will also be allowed on `[ObservableProperty]`:

```diff
-[AttributeUsage(AttributeTargets.Field, AllowMultiple = false, Inherited = false)]
+[AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = false, Inherited = false)]
 public sealed class ObservablePropertyAttribute : Attribute
 {
 }
```

Analyzers will also need to be added to ensure that the annotated properties are partial and valid.

### Usage example

Code like this:

```csharp
[ObservableProperty] 
public partial string Name { get; private set; }
```

Will generate the following:

```csharp
/// <inheritdoc cref="_itemsSource"/>
[global::System.CodeDom.Compiler.GeneratedCode("CommunityToolkit.Mvvm.SourceGenerators.ObservablePropertyGenerator", "8.1.0.0")]
[global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
public partial string Name
{
    get => field;
    set
    {
        if (!global::System.Collections.Generic.EqualityComparer<string>.Default.Equals(field, value))
        {
            OnNameChanging(value);
            OnPropertyChanging(global::CommunityToolkit.Mvvm.ComponentModel.__Internals.__KnownINotifyPropertyChangingArgs.Name);
            
            field = value;

            OnItemsSourceChanged(value);
            OnNameChanged(global::CommunityToolkit.Mvvm.ComponentModel.__Internals.__KnownINotifyPropertyChangedArgs.Name);
        }
    }
}

/// <summary>Executes the logic for when <see cref="Name"/> is changing.</summary>
[global::System.CodeDom.Compiler.GeneratedCode("CommunityToolkit.Mvvm.SourceGenerators.ObservablePropertyGenerator", "8.1.0.0")]
partial void OnNameChanging(string value);

/// <summary>Executes the logic for when <see cref="Name"/> just changed.</summary>
[global::System.CodeDom.Compiler.GeneratedCode("CommunityToolkit.Mvvm.SourceGenerators.ObservablePropertyGenerator", "8.1.0.0")]
partial void OnNameChanged(string value);
```

This will also enable support for declaring custom accessibility modifiers for property accessors.

### Progress tracking

- [X] Initial code generation support
- [X] Add analyzer for C# preview
- [X] Add analyzer to use partial properties instead of fields
- [x] Add code fixer to convert fields to partial properties
- [x] Add analyzer for partial property declarations
- [x] Add analyzer for unsupported Roslyn version for properties
- [x] Update all titles/descriptions of existing shared diagnostics

Sergio0694:
**Closes #555**

This PR updates the `[ObservableProperty]` generator to support partial properties as well.
It includes the following changes:
- Update the target of all attribute types to support properties too
- Update generator to handle partial properties
- Update all diagnostic messages to also mention partial properties
- Add a new analyzer to error if C# is not set to preview
- Add a new analyzer to error if using a version of Roslyn that's too old
- Add a new analyzer to error if a partial property declaration is invalid
- Add a new analyzer to suggest converting to a partial property when possible
- Add a new code fixer to automatically update all fields to partial properties
- Add unit tests for all new features

## PR Checklist


src/CommunityToolkit.Mvvm.SourceGenerators/Diagnostics/Analyzers/InvalidPropertyLevelObservablePropertyAttributeAnalyzer.cs
```diff
@@ -0,0 +1,118 @@
...
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class InvalidPropertyLevelObservablePropertyAttributeAnalyzer : DiagnosticAnalyzer
+{
+    /// <inheritdoc/>
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(InvalidPropertyDeclarationForObservableProperty);
+
+    /// <inheritdoc/>
+    public override void Initialize(AnalysisContext context)
+    {
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
```
dongle-the-gadget, line 29: Couldn't this just be `GeneratedCodeAnalysisFlags.None`?

Sergio0694: Oh, good catch! Not really sure why I had set these to analyzer generated code, none of the analyzers in the MVVM Toolkit should really do that, since they're only meant to detect incorrect uses of the attribute on your own code. I've removed that in all of them, which in theory should also optimize things a bit more. And I've enabled this analyzer for older Roslyn too while at it, since it doesn't need missing APIs anymore 🙂

src/CommunityToolkit.Mvvm/ComponentModel/Attributes/ObservablePropertyAttribute.cs
```diff
@@ -20,38 +20,54 @@ namespace CommunityToolkit.Mvvm.ComponentModel;
 /// partial class MyViewModel : ObservableObject
 /// {
 ///     [ObservableProperty]
-///     private string name;
+///     public partial string name { get; set; }
```
JochemPalmsens, line 23: Just a question: should "name" here also have a capital, since it's now a property? and the generated code below also seems to have a capital?

PR #1120, opened 2025-08-15
fix object.Equals for EquatableArray

### LINKED ISSUE #1119: object.Equals() in EquatableArray{T} causes an infinite loop
### Describe the bug

The source generator project's helper struct `EquatableArray` will cause a stack overflow if `object.Equals()` is called.
This is because `Equals(this, foo)` will call `object.Equals()`, resulting in an infinite loop.
We need to call the typed version from `IEquatable<T>` instead. 

### Regression

_No response_

### Steps to reproduce

Add this test to CommunityToolkit.Mvvm.SourceGenerators `Test_SourceGeneratorsCodegen.cs` to confirm:

```cs
[TestMethod]
public void EquatableArrayTest()
{
    object a = new EquatableArray<string>();
    var b = new EquatableArray<string>();

    Assert.AreEqual(a, b);
}
```

Before the fix, it will cause a stack overflow.

### Expected behavior

You would expect the test to pass without a stack overflow.

### Screenshots

_No response_

### IDE and version

VS 2022

### IDE version

_No response_

### Nuget packages

- [ ] CommunityToolkit.Common
- [x] CommunityToolkit.Mvvm (aka MVVM Toolkit)

### Nuget package version(s)

8.4.0

### Additional context

_No response_

### Help us help you

JamesYFC:
**Closes #1119**

## PR Checklist

- [ ] New component


JamesYFC:
@dotnet-policy-service agree

PR #1162, opened 2026-02-07
Remove explicit backing field generated Relay Commands when possible

### LINKED ISSUE #1073: Use `field` keyword on generated relay commands
### Overview

If `LangVersion=preview`, the relay command generator should avoid emitting a backing field.

### API breakdown

N/A

### Usage example

N/A

### Breaking change?

No

### Alternatives

N/A

### Additional context

_No response_

### Help us help you

Avid29:
**Closes #1073**

When the C# language version supports it, this change replaces the backing field for generated `RelayCommand`s with an implicit field using the new `field` keyword.

## PR Checklist

## Other information

Technically, this is a breaking change since someone may have been using the backing fields for whatever reason

### DIFF
--- a/src/CommunityToolkit.Mvvm.SourceGenerators/Input/RelayCommandGenerator.Execute.cs
+++ b/src/CommunityToolkit.Mvvm.SourceGenerators/Input/RelayCommandGenerator.Execute.cs
@@ -59,7 +59,7 @@ partial class RelayCommandGenerator
             token.ThrowIfCancellationRequested();
 
             // Get the command field and property names
-            (string fieldName, string propertyName) = GetGeneratedFieldAndPropertyNames(methodSymbol);
+            (string? fieldName, string propertyName) = GetGeneratedFieldAndPropertyNames(methodSymbol, semanticModel.Compilation);
 
             token.ThrowIfCancellationRequested();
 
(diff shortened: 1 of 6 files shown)


src/CommunityToolkit.Mvvm.SourceGenerators/Input/RelayCommandGenerator.Execute.cs
```diff
@@ -207,25 +207,33 @@ public static ImmutableArray<MemberDeclarationSyntax> GetSyntax(CommandInfo comm
...
-                .AddModifiers(Token(SyntaxKind.PrivateKeyword))
-                .AddAttributeLists(
-                    AttributeList(SingletonSeparatedList(
-                        Attribute(IdentifierName("global::System.CodeDom.Compiler.GeneratedCode"))
-                        .AddArgumentListArguments(
-                            AttributeArgument(LiteralExpression(SyntaxKind.StringLiteralExpression, Literal(typeof(RelayCommandGenerator).FullName))),
-                            AttributeArgument(LiteralExpression(SyntaxKind.StringLiteralExpression, Literal(typeof(RelayCommandGenerator).Assembly.GetName().Version.ToString()))))))
-                    .WithOpenBracketToken(Token(TriviaList(Comment($"/// <summary>The backing field for <see cref=\"{commandInfo.PropertyName}\"/>.</summary>")), SyntaxKind.OpenBracketToken, TriviaList())))
-                .AddAttributeLists(forwardedFieldAttributes);
+            ImmutableArrayBuilder<MemberDeclarationSyntax> declarations = ImmutableArrayBuilder<MemberDeclarationSyntax>.Rent();
```
huguesv: Based on other usage of `ImmutableArrayBuilder.Rent`, it's meant to be disposed.
https://github.com/search?q=repo%3ACommunityToolkit%2Fdotnet%20ImmutableArrayBuilder&type=code

huguesv:
I'd love for this to go in. I had looked into what it meant to implement this a few months ago and it was pretty much what you ended up doing. Thanks for actually doing it!

I was too wondering what to do about the breaking change to anyone using the field directly. It seems only avoidable if we require folks to opt-in (via an extra parameter to the attribute?) but that would make the optimal scenario more verbose and worse in my opinion.

src/CommunityToolkit.Mvvm.SourceGenerators/Input/RelayCommandGenerator.Execute.cs
Avid29: Nice catch. I thought that the `ToImmutable` call at the end of the function disposed the builder, but upon further investigation, I was wrong. I'll add a using declaration, so dispose is called at the end of the scope.

Avid29:
@huguesv I've added an optional named parameter to the `RelayCommandAttribute` named `ForceBackingField`, which can be used to force a backing field to be used even when the `field` keyword is available.

PR #1191, opened 2026-05-01
Add net10.0 and net10.0-windows10.0.17763.0 target frameworks to CommunityToolkit.Mvvm

Rolling2405:
## Summary

Adds 
et10.0 and 
et10.0-windows10.0.17763.0 to CommunityToolkit.Mvvm's TargetFrameworks, following the existing LTS-only shipping pattern (net8.0 LTS → net10.0 LTS, skipping net9.0 STS).

## Changes

- src/CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.csproj: add 
et10.0;net10.0-windows10.0.17763.0 to TargetFrameworks
- 	ests/CommunityToolkit.Mvvm.Roslyn5000.UnitTests/: add 
et10.0 (was the only MVVM runtime test project missing it; all 408 tests pass)

## Why no other changes are needed

All Windows-conditional property groups use IsTargetFrameworkCompatible(..., 'net8.0-windows10.0.17763.0') which returns 	rue for 
et10.0-windows10.0.17763.0 automatically. All NET8_0_OR_GREATER symbols are true for net10.0, so no source code changes are required. The global.json already pins SDK 10.0.101, and CI uses global.json, so no CI changes are needed.

## Verification

Built locally with dotnet build src/CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.csproj -warnaserror: **0 warnings, 0 errors**, with confirmed output for 
et10.0\CommunityToolkit.Mvvm.dll and 
et10.0-windows10.0.17763.0\CommunityToolkit.Mvvm.dll.

PR #1192, opened 2026-05-01
perf(CommunityToolkit.Mvvm): use System.Threading.Lock (NET9+) and FrozenDictionary (NET8+)

Rolling2405:
## Summary

This PR adds two targeted performance improvements to CommunityToolkit.Mvvm for projects targeting .NET 9+ or .NET 8+, guarded by #if NET9_0_OR_GREATER / #if NET8_0_OR_GREATER so all existing TFMs are completely unaffected.

---

### 1. \System.Threading.Lock\ for messenger lock fields (NET9+)

**Files changed:** \WeakReferenceMessenger.cs\, \StrongReferenceMessenger.cs\, \ConditionalWeakTable2{TKey,TValue}.ZeroAlloc.cs\

Both messengers previously used \lock (this.recipientsMap)\ — locking directly on the data-holding dictionary field. This PR introduces a dedicated \
ecipientsMapLock\ field (typed as \System.Threading.Lock\ on NET9+, \object\ on older targets) and redirects all lock sites to use it.

On .NET 9+, the C# 13 \lock\ statement automatically calls \Lock.EnterScope()\ instead of \Monitor.Enter\/\Monitor.Exit\, which:
- Avoids the atomic compare-exchange inside \Monitor.Enter\'s SyncBlock lookup
- Enables a truly non-blocking \TryEnter\ path (\WeakReferenceMessenger.CleanupWithNonBlockingLock\)
- Separates data ownership from synchronization semantics

\ConditionalWeakTable2\ uses a cross-method lock held across \GetEnumerator()\/\Enumerator.Dispose()\. Because \Lock.Scope\ is a \
ef struct\ it cannot span method boundaries, so this site uses \lockObject.Enter()\/\lockObject.Exit()\ directly with \#if\ guards (fixing the CS9216 boxing-of-Lock diagnostic).

\CommunityToolkit.Mvvm.csproj\ is updated to \LangVersion=13.0\ to enable the optimised lock statement. This mirrors the pattern already used in the Roslyn test projects (\Roslyn5000.UnitTests\ uses 14.0) and only affects this library project.

### 2. \FrozenDictionary\ for \ObservableValidator.DisplayNamesMap\ (NET8+)

**File changed:** \ObservableValidator.cs\

\DisplayNamesMap\ is a \ConditionalWeakTable<Type, Dictionary<string,string>>\ — the inner dictionary is built once per validated type via reflection, then only ever read. On NET8+, this PR changes the inner dictionary type to \FrozenDictionary<string,string>\:
- \FrozenDictionary\ generates a minimal perfect hash at creation time
- All subsequent \TryGetValue\ calls (on every validated property) use constant-time lookups with no collision chains
- The \ConditionalWeakTable\ wrapper handles lifetime correctly (\FrozenDictionary\ is a reference type ✅)

---

## Testing

All 408 tests in \CommunityToolkit.Mvvm.Roslyn5000.UnitTests\ pass on \
et10.0\:

\\\
Passed! - Failed: 0, Passed: 408, Skipped: 0, Total: 408
\\\

The \SourceGenerators.Roslyn5000.UnitTests\ project was intentionally not targeted — it has pre-existing failures on \
et10.0\ unrelated to this change (diagnostic text differences due to C# 14 defaults), as noted in PR #1191.

---

## Related

Companion to PR #1191 (net10.0 TFM addition). These improvements are meaningful regardless of that PR but are most impactful when the library is consumed on .NET 9/10.

### DIFF
--- a/src/CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.csproj
+++ b/src/CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netstandard2.1;net8.0;net8.0-windows10.0.17763.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net8.0;net8.0-windows10.0.17763.0;net10.0;net10.0-windows10.0.17763.0</TargetFrameworks>
   </PropertyGroup>
 
   <!--
--- a/src/CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.csproj
+++ b/src/CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.csproj
@@ -2,6 +2,9 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;netstandard2.1;net8.0;net8.0-windows10.0.17763.0;net10.0;net10.0-windows10.0.17763.0</TargetFrameworks>
+    <!-- C# 13 enables the optimized lock statement for System.Threading.Lock (NET9+), where
+         the compiler emits Lock.EnterScope() instead of Monitor.Enter/Exit for reduced overhead. -->
+    <LangVersion>13.0</LangVersion>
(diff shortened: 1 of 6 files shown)


Rolling2405:
@dotnet-policy-service agree

PR #1207, opened 2026-07-26
Fix CanExecute(null) for value types without predicate

### LINKED ISSUE #1206: RelayCommand<T>.CanExecute(null) returns false for value types without a predicate
### Describe the bug

`RelayCommand<T>.CanExecute(object?)` unconditionally returns `false` when `T` is a non-nullable value type (e.g. `int`, enums) and the `parameter` is `null` — even when no `canExecute` predicate was provided to the constructor.

This violates the documented contract: *"The default return value for the CanExecute method is `true`."* A command constructed without a predicate is meant to always be executable, regardless of the parameter value — including `null`.

In WPF applications on .NET Framework, the framework may call `CanExecute(null)` during control initialization before the `CommandParameter` binding has produced a value. With the unconditional `return false`, commands without predicates (e.g. `RelayCommand<SomeEnum>(DoSomething)`) are permanently disabled.

**Current code** (`src/CommunityToolkit.Mvvm/Input/RelayCommand{T}.cs`):

```csharp
if (parameter is null && default(T) is not null)
{
    return false;   // unconditional — even when no predicate exists
}
```

### Regression

_No response_

### Steps to reproduce

```csharp
// 1. Create a RelayCommand without a predicate
var command = new RelayCommand<int>(i => Console.WriteLine(i));

// 2. Call CanExecute with null (as WPF does during initialization)
bool result = ((System.Windows.Input.ICommand)command).CanExecute(null);

// 3. Observe the result
Console.WriteLine(result); // Actual: False, Expected: True
```

### Expected behavior

`CanExecute(null)` should return `true` when no `canExecute` predicate was provided, because:
1. The class documentation states the default return value is `true`
2. Without a predicate, the command has no condition to evaluate — it should always be executable
3. The `null` guard's purpose is to prevent a `null` from being incorrectly mapped to `default(T)` (e.g., 0 as SelectedIndex), but when there's no predicate, there's nothing to incorrectly map to

For commands *with* a predicate, returning `false` is correct — we cannot guess the user's intent. This proposal keeps that behavior unchanged.

### Screenshots

_No response_

### IDE and version

Other

### IDE version

Insiders [12020.428]

### Nuget packages

- [ ] CommunityToolkit.Common
- [x] CommunityToolkit.Mvvm (aka MVVM Toolkit)

### Nuget package version(s)

latest stable

### Additional context

- Related discussion: [WindowsCommunityToolkit #3619](https://github.com/CommunityToolkit/WindowsCommunityToolkit/issues/3619). That issue correctly identified that mapping `null` to `default(T)` for commands *with* predicates is unsafe. However, the resulting unconditional `return false` is overly broad — it also disables commands without predicates, which have no such ambiguity.
- The proposed fix is a one-line change: `return false;` → `return this.canExecute is null;`
- `TryGetCommandArgument`、`ThrowArgumentExceptionForInvalidCommandArgument`、`Execute(object?)` remain unchanged.

### Help us help you

apachezy:
Closes #1206

When `RelayCommand<T>.CanExecute(object?)` receives `null` for a non-nullable value type `T`, it unconditionally returns `false` — even when no `canExecute` predicate was provided. This violates the documented contract: *"The default return value for the CanExecute method is `true`."*

This single-line fix distinguishes between the two cases:

- **No predicate** → return `true` (the command is always executable, as documented)
- **Has predicate** → return `false` (cannot meaningfully evaluate `null` as `T`, safe fallback — unchanged from current behavior)

In WPF applications on .NET Framework, the framework may call `CanExecute(null)` during control initialization before the `CommandParameter` binding has produced a value. With the unconditional `return false`, commands without predicates are permanently disabled, as the initial `false` result is cached and never re-evaluated.

This fix is fully compatible with the reasoning in [WindowsCommunityToolkit #3619](https://github.com/CommunityToolkit/WindowsCommunityToolkit/issues/3619): for commands *with* a predicate, mapping `null` to `default(T)` would conflate "no value" with a legitimate value (e.g., `0` as `SelectedIndex`). The conclusion that `return false` is safe for that scenario is preserved here. The case of *no predicate* was not explicitly considered in that discussion — a command without a predicate should simply always be executable, as the class documentation states.

## PR Checklist

- [ ] Created a feature/dev branch in your fork (vs. submitting directly from a commit on main)
- [ ] Based off latest main branch of toolkit
- [ ] PR doesn't include merge commits (always rebase on top of our main, if needed)
- [ ] New component

## Other information

**Change in `src/CommunityToolkit.Mvvm/Input/RelayCommand{T}.cs`:**

```diff
 if (parameter is null && default(T) is not null)
 {
-    return false;
+    return this.canExecute is null;
 }
```

| Scenario | `canExecute` | Before | After |
|---|---|---|---|
| `RelayCommand<int>(Do)` | `null` | `false` ❌ | `true` ✅ |
| `RelayCommand<int>(Do, i => i >= 0)` | predicate | `false` | `false` (unchanged) |

`TryGetCommandArgument`、`ThrowArgumentExceptionForInvalidCommandArgument`、`Execute(object?)` 均未改动。

### DIFF
--- a/src/CommunityToolkit.Mvvm/Input/RelayCommand{T}.cs
+++ b/src/CommunityToolkit.Mvvm/Input/RelayCommand{T}.cs
@@ -82,11 +82,20 @@ public sealed partial class RelayCommand<T> : IRelayCommand<T>
     /// <inheritdoc/>
     public bool CanExecute(object? parameter)
     {
-        // Special case a null value for a value type argument type.
-        // This ensures that no exceptions are thrown during initialization.
+        // WPF may call CanExecute(null) during initialization before the
+        // CommandParameter binding has produced a value. For non-nullable
+        // value types (e.g. int, enums), null cannot be meaningfully
+        // converted — there is no "default" that can safely stand in for
+        // "no value" (0 could be a legitimate SelectedIndex, for example).
+        //
+        // - If the user didn't supply a canExecute predicate, the command
+        //   is meant to always run — return true.
+        // - If a predicate exists, we cannot guess the user's intent, so
+        //   return false. The framework will re-query CanExecute when the
+        //   binding eventually produces the real parameter value.
         if (parameter is null && default(T) is not null)
         {
-            return false;
+            return _canExecute is null;
         }
 
         if (!TryGetCommandArgument(parameter, out T? result))
--- a/src/CommunityToolkit.Mvvm/Input/RelayCommand{T}.cs
+++ b/src/CommunityToolkit.Mvvm/Input/RelayCommand{T}.cs
@@ -82,17 +82,16 @@ public sealed partial class RelayCommand<T> : IRelayCommand<T>
     /// <inheritdoc/>
     public bool CanExecute(object? parameter)
     {
-        // WPF may call CanExecute(null) during initialization before the
-        // CommandParameter binding has produced a value. For non-nullable
-        // value types (e.g. int, enums), null cannot be meaningfully
-        // converted — there is no "default" that can safely stand in for
-        // "no value" (0 could be a legitimate SelectedIndex, for example).
-        //
-        // - If the user didn't supply a canExecute predicate, the command
-        //   is meant to always run — return true.
-        // - If a predicate exists, we cannot guess the user's intent, so
-        //   return false. The framework will re-query CanExecute when the
-        //   binding eventually produces the real parameter value.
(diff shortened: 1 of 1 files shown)


apachezy:
@dotnet-policy-service agree [company="None"]
@dotnet-policy-service agree
@dotnet-policy-service agree company="Microsoft"

apachezy:
@dotnet-policy-service agree company="no company"

